PR #2, opened 2020-10-25
Add a readme

ColdIce1605:
Cause why not :P
if you want anything else added I can do that.


coderbot16 (approved):
Seems good for now, thanks! I'll probably write a more detailed README at some point though.

PR #3, opened 2020-10-25
Add the lightmap to the terrain shader

ColdIce1605:
fixed lighting 
 - terrain matches entities
 - terrain now has lighting
@coderbot16

### DIFF
--- a/src/main/resources/gbuffers_textured.fsh
+++ b/src/main/resources/gbuffers_textured.fsh
@@ -1,10 +1,15 @@
 #version 120
 
 uniform sampler2D terrain;
+uniform sampler2D lightmap;
 
 varying vec4 color;
 varying vec2 tex;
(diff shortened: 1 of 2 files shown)


ColdIce1605:
thing to notice the lmcoord doesn't match the tex or to most devs texcoord this doesn't really affect anything but uniformity.

coderbot16 (changes requested):
Looks good, thanks! But if texcoord is more standard then it would probably be good to use that instead.

Also, no need to mention me, I'm watching the repository so I get notifications for everything.

ColdIce1605 (commented):
all changed

ColdIce1605:
another thing is I changed what is assigned to texcoord to be more like lmcoord

coderbot16 (approved):
LGTM

PR #5, opened 2020-10-26
adds viewHeight and viewWidth

ColdIce1605:
I'm getting better at this it seems.
:)

### DIFF
--- a/src/main/java/net/coderbot/iris/uniforms/Uniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/Uniforms.java
@@ -1,17 +1,16 @@
 package net.coderbot.iris.uniforms;
 
-import java.nio.FloatBuffer;
-
 import net.coderbot.iris.texunits.TextureUnit;
-import org.lwjgl.BufferUtils;
-import org.lwjgl.opengl.GL21;
-
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.gl.GlProgram;
 import net.minecraft.client.util.math.Vector3f;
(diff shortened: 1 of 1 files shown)


ColdIce1605:
lmma blame this on my new IDE and not me.

ColdIce1605:
hmm hold off something not right

src/main/java/net/coderbot/iris/uniforms/Uniforms.java
```diff
@@ -37,6 +39,12 @@ public Uniforms(GlProgram program) {
 		gbufferProjection = GL21.glGetUniformLocation(programId, "gbufferProjection");
 		gbufferProjectionInverse = GL21.glGetUniformLocation(programId, "gbufferProjectionInverse");
 
+		viewHeight  = MinecraftClient.getInstance().getWindow().getHeight();
```
coderbot16, line 42: This isn't quite right. `private int viewHeight` as declared above actually holds the location of the viewHeight uniform used for sending values to the shaders. I have a few ideas on how to refactor the uniform code to be a bit more intuitive, since right now it's a very basic implementation.

This specific call should probably be happening in a `glUniform*` call within `update()`.

coderbot16 (changes requested):

ColdIce1605:
Thanks I knew something wasn't right

ColdIce1605:
Ok, I've solved it.

ColdIce1605:
and im also adding more uniform but in different PRs

coderbot16 (approved):
Thanks!

PR #7, opened 2020-11-05
Add many more uniforms

ColdIce1605:
Adds 
- eyeAltitude, isEyeInWater, moonPhase, and hideGUI 

Things to note:
Maybe we should group uniforms together by type?
isEyeInWater might be incorrect

### DIFF
--- a/src/main/java/net/coderbot/iris/uniforms/Uniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/Uniforms.java
@@ -25,8 +25,14 @@ public class Uniforms {
 	private int viewWidth;
 
 	private int cameraPosition;
+	
+	private int eyeAltitude;
+	private int isEyeInWater;
 
 	private int shadowLightPosition;
+	private static int moonPhase;
+	
+	private static int hideGUI;
 
 	public Uniforms(GlProgram program) {
 		int programId = program.getProgramRef();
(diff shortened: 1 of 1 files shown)


src/main/java/net/coderbot/iris/uniforms/Uniforms.java
```diff
@@ -62,6 +74,19 @@ public void update() {
 
 
 		updateVector(cameraPosition, MinecraftClient.getInstance().gameRenderer.getCamera().getPos());
+		
+		GL21.glUniform1f(eyeAltitude, (float)MinecraftClient.getInstance().player.getPos().getY());
+		
+		// TODO: Simplify 
+		// Help is it player or cameraEntity
+		int EyeInWater = 0;
+		GL21.glUniform1f(eyeAltitude, (float)MinecraftClient.getInstance().player.getPos().getY());
```
coderbot16, line 83: These should probably be the camera entity

src/main/java/net/coderbot/iris/uniforms/Uniforms.java
```diff
@@ -62,6 +74,19 @@ public void update() {
 
 
 		updateVector(cameraPosition, MinecraftClient.getInstance().gameRenderer.getCamera().getPos());
+		
+		GL21.glUniform1f(eyeAltitude, (float)MinecraftClient.getInstance().player.getPos().getY());
+		
+		// TODO: Simplify 
+		// Help is it player or cameraEntity
+		int EyeInWater = 0;
```
coderbot16, line 82: java naming conventions
```suggestion
		int eyeInWater = 0;
```

coderbot16 (changes requested):
A few small changes. It is also preferable (but not required) to have the implemented uniforms documented. Though, for simpler uniforms, it isn't strictly needed.

Good documentation would help make it clear whether the positions refer to the camera entity or player entity, for example.

coderbot16:
Thanks!

> Maybe we should group uniforms together by type?

Definitely, I'd like to do this once we have a better idea of what the best way to categorize the uniforms is.

PR #13, opened 2020-11-11
add a ci

ColdIce1605:
we can also add a workflow that releases jar files to either to curseforge or just github
also I'm good with CIs(thats a joke).


WalkerKnapp:
Why does there need to be actions for both windows and ubuntu? Shouldn't just one be sufficient for this?

ColdIce1605:
Yeah it should but I'm good with CIs(thats a joke)(I couldn't figure it out[x, x] doesn't work

.github/workflows/windows.yml
```diff
@@ -0,0 +1,26 @@
+# This workflow will build a Java project with Gradle
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Java CI with Gradle
```
coderbot16, line 4: I don't think that testing windows compiles on every push / PR is needed. Also, it looks like the script uses linux commands anyway, so I don't think it will even work on Windows? This file should probably be removed.

coderbot16 (changes requested):
Seems simple enough, but the windows workflow is unnecessary.

.github/workflows/windows.yml
ColdIce1605: it does work anyway ill remove it because java isn't a compiled lang.

coderbot16 (approved):
Thanks!

PR #8, opened 2020-11-07
Add External Shaderpack support and add cloud/block outline shaders

OverlordsIII:
This pr adds external shaderpack support and adds a couple of new shaders for clouds and block outlines.
 * If this pr is accepted, when you start up minecraft, Iris will crash asking you to change your shaderpack in `config/iris.properties`. Simply change the end value of `shaderpack` to be the unzipped folder of the shaderpack. Make sure that the unzipped shaderpack has the `shader` directory directly inside of it.
* It also adds shaders for clouds and block outlines (basic)
* adds parsing of `item.properties` files and `block.properties` files in shaderpacks. Will be used later for uniforms and attributes. Can return the item and block map by calling `Iris.getShaderParser().getItemProperties()` or `Iris.getShaderParser().getBlockProperties()`

I probably messed something up, so please tell me what is wrong so I can fix it.
Thanks for letting me pr on this project.

PR #15, opened 2020-11-16

OverlordsIII:
Just a rewrite of what part of #8 was. Adds parsing for blocks, items, and entities. Adds uniforms `heldItemId`, `heldItemid2`, `blockEntityId`, and `entityId`


coderbot16:
I started reviewing this PR but noticed that the git history is really messed up here. It looks like the last two commits are the only relevant ones to the PR. I squash-merge PRs, so you should always make a new branch off of the current `master` before beginning work on the PR. I can still review this PR, but it would be preferable if you could fix up the git history.

OverlordsIII:
oops sorry about that lemme make a new branch
I think I have to close this then 
sry about that

PR #16, opened 2020-11-16

OverlordsIII:
same as #15 with (hopefully) better git history and a new branch


coderbot16:
Appears to have the same problems as #15. You could re-clone in a separate directory, delete .git for your current project directory, and then drag the .git from the new clone to your current project directory, then commit your changes as usual. That should fix the history. In the future, make a new branch from `master`, and only *then* begin committing changes for the new PR.

PR #14, opened 2020-11-16
change all prints to log

ColdIce1605:
nice :)

PR #17, opened 2020-11-16

OverlordsIII:
Does the same thing as #15 . Hopefully git history works. I did as you said and deleted `.git`, so its back to just one commit

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/AccessorRenderLayers.java
+++ b/src/main/java/net/coderbot/iris/mixin/AccessorRenderLayers.java
@@ -1,0 +1,17 @@
+package net.coderbot.iris.mixin;
+
+import net.minecraft.block.Block;
+import net.minecraft.client.render.RenderLayer;
+import net.minecraft.client.render.RenderLayers;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import java.util.Map;
+
+@Mixin(RenderLayers.class)
+public interface AccessorRenderLayers {
+    @Accessor("BLOCKS")
+    static Map<Block, RenderLayer> getBlockRenderLayers(){
+        throw new AssertionError("Accessor Mixin \"AccessorRenderLayers\" Failed to apply!");
+    }
+}
--- a/src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
+++ b/src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
@@ -1,25 +1,31 @@
 
 package net.coderbot.iris.mixin;
 
+import it.unimi.dsi.fastutil.objects.ObjectListIterator;
 import net.coderbot.iris.HorizonRenderer;
 import net.coderbot.iris.Iris;
 import net.coderbot.iris.uniforms.CapturedRenderingState;
+import net.minecraft.block.entity.BlockEntity;
+import net.minecraft.client.render.*;
+import net.minecraft.entity.Entity;
+import net.minecraft.util.math.Vec3d;
+import net.minecraft.util.profiler.Profiler;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.Slice;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
-import net.minecraft.client.render.Camera;
-import net.minecraft.client.render.GameRenderer;
-import net.minecraft.client.render.LightmapTextureManager;
-import net.minecraft.client.render.RenderLayer;
-import net.minecraft.client.render.WorldRenderer;
 import net.minecraft.client.util.math.MatrixStack;
 import net.minecraft.util.math.Matrix4f;
 
 import net.fabricmc.api.EnvType;
 import net.fabricmc.api.Environment;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
 
 @Mixin(WorldRenderer.class)
 @Environment(EnvType.CLIENT)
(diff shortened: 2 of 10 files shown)


OverlordsIII:
also I added a `println` in IdMapParser. I'll change it when #14 gets approved

ColdIce1605:
Now that's it's in scope add shader.properties?
That includes 
Option parsing
Custom uniforms
And setting different images to textures(required some sampler2D uniforms tho)

coderbot16:
Most of the things in shader.properties each probably deserve separate PRs, I wouldn't say that they're in scope of this PR. The sweet spot for the size of a PR is 200-400 lines, and this one is 413 lines.

In general I much prefer smaller PRs, since it's incredibly exhausting to review a 1000+ line PR for example. This PR is fine as is, size-wise.

src/main/java/net/coderbot/iris/mixin/AccessorRenderLayers.java
```diff
@@ -0,0 +1,17 @@
...
+import net.minecraft.block.Block;
+import net.minecraft.client.render.RenderLayer;
+import net.minecraft.client.render.RenderLayers;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import java.util.Map;
+
+@Mixin(RenderLayers.class)
+public interface AccessorRenderLayers {
```
coderbot16, line 12: Per my comment in IdMapParser, this system as-is is too fragile. I'm removing it for now.

src/main/java/net/coderbot/iris/shaderpack/IdMapParser.java
```diff
@@ -0,0 +1,176 @@
...
+     * A map that contains the entity ids of an entity to the int value
+     */
+    private Map<Identifier, Integer> entityPropertiesMap = Maps.newHashMap();
+
+    IdMapParser(Path shaderPath){
+        this.shaderPath = shaderPath;
+        parseBlockProperties();
+        parseItemProperties();
+        parseEntityProperties();
+        this.blockRenderLayerMap.forEach((identifier, layer) -> AccessorRenderLayers.getBlockRenderLayers().put(Registry.BLOCK.get(identifier), layer));
```
coderbot16, line 44: This only works properly if the user never changes the shaderpack while the game is running. I'm removing it for now, but in the future a better system will probably be necessary.

src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
```diff
@@ -112,4 +118,15 @@
...
+
+	@Inject(method = "renderEntity", at = @At("HEAD"))
+	private void iris$getCurrentEntity(Entity entity, double cameraX, double cameraY, double cameraZ, float tickDelta, MatrixStack matrices, VertexConsumerProvider vertexConsumers, CallbackInfo ci){
+		CapturedRenderingState.INSTANCE.setCurrentEntity(entity);
+	}
+
+	@Inject(method = RENDER, at = @At(value = "INVOKE", target = "Lnet/minecraft/block/entity/BlockEntity;getPos()Lnet/minecraft/util/math/BlockPos;", ordinal = 1), locals = LocalCapture.CAPTURE_FAILHARD)
+	private void iris$getCurrentBlockEntity(MatrixStack matrices, float tickDelta, long limitTime, boolean renderBlockOutline, Camera camera, GameRenderer gameRenderer, LightmapTextureManager lightmapTextureManager, Matrix4f matrix4f, CallbackInfo ci, Profiler profiler, Vec3d vec3d, double d, double e, double f, Matrix4f matrix4f2, boolean bl, Frustum frustum2, boolean bl3, VertexConsumerProvider.Immediate immediate, Set var39, Iterator var40, BlockEntity blockEntity2){
+		CapturedRenderingState.INSTANCE.setCurrentBlockEntity(blockEntity2);
+	}
```
coderbot16, line 130: These aren't properly cleared when entities *aren't*  being rendered, so the uniforms will hold unspecified values. It would probably be better for them to just hold `null`.

src/main/java/net/coderbot/iris/pipeline/ShaderPipeline.java
```diff
@@ -48,8 +50,13 @@ public ShaderPipeline(ShaderPack pack) {
 		this.terrain = texturedLit;
 		this.translucent = terrain;
 		this.weather = texturedLit;
+		this.pack = pack;
 	}
 
+	public ShaderPack getPack() {
```
coderbot16, line 56: The pipeline shouldn't hold on to the source shader pack, IMO.

src/main/java/net/coderbot/iris/uniforms/PropertiesUniforms.java
```diff
@@ -0,0 +1,82 @@
...
+    private static IntSupplier getHeldItemId(Hand hand) {
+        if (MinecraftClient.getInstance().player != null) {
+            Identifier currentStack = Registry.ITEM.getId(MinecraftClient.getInstance().player.getStackInHand(hand).getItem());
+            IdMapParser parser = Iris.getPipeline().getPack().getIdMapParser();
+            if (parser.getItemProperties().containsKey(currentStack)) {
+                return () -> parser.getItemProperties().get(currentStack);
+            }
+        }
+        return () -> -1;
+    }
```
coderbot16, line 46: This isn't how uniform attachments work. `getHeldItemId` will be called exactly one time for a given shader pack when the shader pipeline is compiled, and then never again. This will cause subtle bugs depending on what item the player is holding at the time that the shader pipeline is initialized.

src/main/java/net/coderbot/iris/uniforms/PropertiesUniforms.java
```diff
@@ -0,0 +1,82 @@
...
+import net.minecraft.client.render.block.entity.BlockEntityRenderDispatcher;
+import net.minecraft.entity.Entity;
+import net.minecraft.util.Hand;
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.Registry;
+
+import java.util.Map;
+import java.util.function.IntSupplier;
+
+public final class PropertiesUniforms {
```
coderbot16, line 19: I don't like the name of this class, it's a bit ambiguous since it's possible to define custom uniforms in `shaders.properties` as well.

src/main/resources/mixins.iris.json
```diff
@@ -4,16 +4,17 @@
   "package": "net.coderbot.iris.mixin",
   "compatibilityLevel": "JAVA_8",
   "client": [
```
coderbot16, line 6: The indentation changes here should be rolled back

src/main/java/net/coderbot/iris/shaderpack/IdMapParser.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Parses entries from entity.properties into the entityProperties map
+     */
+    private void parseEntityProperties(){
+        Properties properties = new Properties();
+        try {
+            properties.load(Files.newInputStream(shaderPath.resolve("entity.properties")));
+        } catch (IOException e) {
+            System.out.println("An entity.properties was not found in the current shaderpack");
```
coderbot16, line 55: These probably shouldn't be printed to System.out, honestly. Maybe a debug message at most

coderbot16 (changes requested):
Here are the issues I noticed, I'm going to fix them locally myself though.

ColdIce1605:
Also something to note is that optifine parses blockstates and when I was testing these on packs it doesn't like that

coderbot16:
I merged this locally but GitHub didn't recognize it. Oh well.

PR #21, opened 2020-11-30
Add Reload Shaders Keybind and language data parsing

OverlordsIII:
Adds a reload shaders keybind `R` and adds lang entries from shaderpacks and Iris itself into minecraft's language cache


src/main/java/net/coderbot/iris/mixin/lang/MixinTranslationStorage.java
```diff
@@ -0,0 +1,39 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
+
+import java.io.InputStream;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+@Mixin(TranslationStorage.class)
+public class MixinTranslationStorage {
```
coderbot16, line 19: It would probably be better to just pull in `fabric-resource-loader` instead of doing this

src/main/java/net/coderbot/iris/pipeline/ShaderPipeline.java
```diff
@@ -18,25 +18,29 @@
  */
 public class ShaderPipeline {
```
coderbot16, line 19: The fields here should stay final, and you should just create a new ShaderPipeline instance instead externally.

src/main/java/net/coderbot/iris/mixin/keybind/MixinGameOptions.java
```diff
@@ -0,0 +1,29 @@
...
+import org.spongepowered.asm.mixin.Mutable;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.List;
+
+@Mixin(GameOptions.class)
+public class MixinGameOptions {
```
coderbot16, line 18: Fabric-api has a fabric-keybindings module, we should use that instead.

src/main/java/net/coderbot/iris/mixin/keybind/MixinKeybinding.java
```diff
@@ -0,0 +1,27 @@
...
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.util.Map;
+
+@Mixin(KeyBinding.class)
```
coderbot16, line 13: Ditto

src/main/java/net/coderbot/iris/mixin/keybind/MixinMinecraftClient.java
```diff
@@ -0,0 +1,30 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import java.io.IOException;
+
+//just for keybinds, we could move this to a separate class if needed
+@Mixin(MinecraftClient.class)
+public class MixinMinecraftClient {
```
coderbot16, line 15: Ditto

src/main/resources/assets/iris/lang/en_us.json
```diff
@@ -0,0 +1,5 @@
+{
+  "iris.shaders.reloaded": "Reloaded Shaders!",
+  "iris.keybind.reload": "Reload Shaders",
+  "iris.keybinds": "Iris Keybinds"
```
coderbot16, line 4: I think the convention is to not have a "Keybinds" part, so just an "Iris" section instead.

src/main/java/net/coderbot/iris/mixin/lang/MixinTranslationStorage.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private static final String LANGUAGE_LOAD = "Lnet/minecraft/util/Language;load(Ljava/io/InputStream;Ljava/util/function/BiConsumer;)V";
+
+    @Inject(method = "load(Ljava/util/List;Ljava/util/Map;)V", at = @At(value = "INVOKE", target = LANGUAGE_LOAD), locals = LocalCapture.CAPTURE_FAILHARD)
+    private static void iris$addLangEntries(List resources, Map<String, String> translationMap, CallbackInfo ci, Iterator var2, Resource resource, InputStream inputStream) {
+        String fileName = resource.getId().getPath().substring(resource.getId().getPath().indexOf("/"));
+        String langCode = fileName.substring(1, fileName.lastIndexOf("."));
+        try {
+            Language.load(Iris.class.getResourceAsStream("/assets/iris/lang" + fileName), translationMap::put);
+            Map<String, Map<String, String>> langMap = Iris.getCurrentPack().getLangMap();
```
coderbot16, line 29: This isn't really the best approach. You insert values into the language map from the current shaderpack, but when you load a different shaderpack, the values from the previous shaderpack remain. Instead, it would be better to just allow all queries against the translationMap to optionally access this secondary language map instead. Or, maybe have a secondary instance of TranslationStorage. There just needs to be a way that doesn't pollute the vanilla language map with values from the shaderpack language files.

Alternatively, this *could* work as long as you force a resource reload upon loading a different shaderpack, but that seems like a bit overkill at the moment

src/main/java/net/coderbot/iris/mixin/lang/MixinTranslationStorage.java
```diff
@@ -0,0 +1,39 @@
...
+            if (langMap.containsKey(langCode)){
+                translationMap.putAll(langMap.get(langCode));
+            }
+            System.out.println(translationMap);
```
coderbot16, line 33: Log spam
```suggestion
```

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -100,6 +105,22 @@ public static ShaderPipeline getPipeline() {
...
+	public static void reload() throws IOException {
+		//currently this first line can be used to reload to a diff shaderpack, but it should be removed
+		//when there is a gui or a better system for changing packs at runtime
+		//or could be kept for the gui to use
+		irisConfig.initialize();
+		Path shaderpacksDirectory = FabricLoader.getInstance().getGameDir().resolve("shaderpacks");
+		Path shaderPackRoot = shaderpacksDirectory.resolve(irisConfig.getShaderPackName());
+		Path shaderPackPath = shaderPackRoot.resolve("shaders");
+		currentPack.reload(shaderPackPath);
+		pipeline.reload(currentPack);
```
coderbot16, line 117: This should create new ShaderPack and ShaderPipeline instances so that they can remain immutable internally.

This will also crash with a NullPointerException if the ShaderPipeline isn't actually created yet.

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -100,6 +105,22 @@ public static ShaderPipeline getPipeline() {
...
 	}
 
+	public static void reload() throws IOException {
+		//currently this first line can be used to reload to a diff shaderpack, but it should be removed
+		//when there is a gui or a better system for changing packs at runtime
+		//or could be kept for the gui to use
+		irisConfig.initialize();
+		Path shaderpacksDirectory = FabricLoader.getInstance().getGameDir().resolve("shaderpacks");
+		Path shaderPackRoot = shaderpacksDirectory.resolve(irisConfig.getShaderPackName());
+		Path shaderPackPath = shaderPackRoot.resolve("shaders");
```
coderbot16, line 115: It feels like there's probably some code duplication going on here with the main Iris initialization routines

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -1,28 +1,38 @@
...
-	private final ProgramSource gbuffersSkyBasic;
-	private final ProgramSource gbuffersSkyTextured;
-	private final ProgramSource gbuffersClouds;
-	private final IdMap idMap;
+	private ProgramSource gbuffersBasic;
+	private ProgramSource gbuffersTextured;
+	private ProgramSource gbuffersSkyBasic;
+	private ProgramSource gbuffersSkyTextured;
+	private ProgramSource gbuffersClouds;
+	private IdMap idMap;
```
coderbot16, line 18: These should stay immutable

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -89,6 +102,27 @@ private static String readFile(Path path) throws IOException {
...
+					properties.load(Files.newInputStream(path));
+				} catch (IOException e) {
+					e.printStackTrace();
+				}
+				properties.forEach((key, value) -> currentLanguageMap.put(key.toString(), value.toString()));
+				allLanguagesMap.put(currentLangCode, currentLanguageMap);
+			});
+		}
+		return allLanguagesMap;
+	}
```
coderbot16, line 124: It's hard to read this function since there's absolutely zero blank lines

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -89,6 +102,27 @@ private static String readFile(Path path) throws IOException {
...
+		if (Files.exists(langFolderPath)) {
+			Files.walk(langFolderPath, 1).filter(path -> !Files.isDirectory(path)).forEach(path -> {
+				Map<String, String> currentLanguageMap = Maps.newHashMap();
+				String currentFileName = path.getFileName().toString().toLowerCase();
+				String currentLangCode = currentFileName.substring(0, currentFileName.lastIndexOf("."));
+				Properties properties = new Properties();
+				try {
+					properties.load(Files.newInputStream(path));
+				} catch (IOException e) {
+					e.printStackTrace();
```
coderbot16, line 117: Should probably print some additional context here.

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -89,6 +102,27 @@ private static String readFile(Path path) throws IOException {
 		}
 	}
 
+	private Map<String, Map<String, String>> parseLangEntries(Path root) throws IOException {
+		Path langFolderPath = root.resolve("lang");
+		Map<String, Map<String, String>> allLanguagesMap = Maps.newHashMap();
+		if (Files.exists(langFolderPath)) {
```
coderbot16, line 108: Prefer an early-return here with an empty `Map`

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -89,6 +102,27 @@ private static String readFile(Path path) throws IOException {
 		}
 	}
 
+	private Map<String, Map<String, String>> parseLangEntries(Path root) throws IOException {
+		Path langFolderPath = root.resolve("lang");
+		Map<String, Map<String, String>> allLanguagesMap = Maps.newHashMap();
```
coderbot16, line 107:
```suggestion
		Map<String, Map<String, String>> allLanguagesMap = new HashMap<>();
```

coderbot16 (changes requested):

OverlordsIII:
I tried to fix all the changes requested, could you take a look at it?

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -100,6 +105,21 @@ public static ShaderPipeline getPipeline() {
 		return pipeline;
 	}
 
+	public static void reload() throws IOException {
+		//currently this first line can be used to reload to a diff shaderpack, but it should be removed
+		//when there is a gui or a better system for changing packs at runtime
+		//or could be kept for the gui to use
+		irisConfig.initialize();
+		Path shaderPackRoot = shaderpacksDirectory.resolve(irisConfig.getShaderPackName());
```
coderbot16, line 113: What if the internal shader pack is loaded? Should probably record which shaderpack is being used somewhere during startup instead of assuming that it's an external one

I know that reloading internal shaders doesn't make a ton of sense but it shouldn't just fail hard

src/main/java/net/coderbot/iris/mixin/MixinMinecraftClient.java
```diff
@@ -0,0 +1,51 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(MinecraftClient.class)
+public abstract class MixinMinecraftClient {
+    @Shadow @Nullable public ClientPlayerEntity player;
+
+    @Shadow @Final public WorldRenderer worldRenderer;
+
+    @Inject(method = "tick", at = @At("TAIL"))
+    private void iris$tickKeybind(CallbackInfo ci) {
```
coderbot16, line 24: `fabric-lifecycle-events-v1` provides a ClientTickCallback, that would probably be a better choice here.

src/main/java/net/coderbot/iris/mixin/MixinMinecraftClient.java
```diff
@@ -0,0 +1,51 @@
...
+        while (Iris.reloadKeybind.wasPressed()){
+            try {
+                Iris.reload();
+                this.worldRenderer.reload();
+                if (this.player != null){
+                    this.player.sendMessage(new TranslatableText("iris.shaders.reloaded"), false);
+                }
+            } catch (Exception e) {
+                Iris.logger.error("Error while reloading Shaders for Iris!", e);
```
coderbot16, line 33:
```suggestion
                Iris.logger.error("Error reloading shaderpack!", e);
```

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -89,6 +98,35 @@ private static String readFile(Path path) throws IOException {
...
+				try {
+					properties.load(Files.newInputStream(path));
+				} catch (IOException e) {
+					Iris.logger.error("Error while parsing languages for shaderpacks!", e);
+				}
+
+				properties.forEach((key, value) -> currentLanguageMap.put(key.toString(), value.toString()));
+
+				allLanguagesMap.put(currentLangCode, currentLanguageMap);
+			});
```
coderbot16, line 125: The indentation is a bit messed up here, I think it's indented one extra level

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -89,6 +98,35 @@ private static String readFile(Path path) throws IOException {
...
 	}
 
+	private Map<String, Map<String, String>> parseLangEntries(Path root) throws IOException {
+		Path langFolderPath = root.resolve("lang");
+		Map<String, Map<String, String>> allLanguagesMap = new HashMap<>();
+
+		if (!Files.exists(langFolderPath)) {
+			return allLanguagesMap;
+		}
+			Files.walk(langFolderPath, 1).filter(path -> !Files.isDirectory(path)).forEach(path -> {
```
coderbot16, line 108: Maybe comment that this 1 parameter is the max depth and why you're choosing a max depth of 1, to make it easier for others to understand the code

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -89,6 +98,35 @@ private static String readFile(Path path) throws IOException {
...
+				Map<String, String> currentLanguageMap = Maps.newHashMap();
+				String currentFileName = path.getFileName().toString().toLowerCase();
+
+				String currentLangCode = currentFileName.substring(0, currentFileName.lastIndexOf("."));
+				Properties properties = new Properties();
+
+				try {
+					properties.load(Files.newInputStream(path));
+				} catch (IOException e) {
+					Iris.logger.error("Error while parsing languages for shaderpacks!", e);
```
coderbot16, line 119: Maybe include the file path in the error message? That would probably be helpful for shaderpack authors

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -89,6 +98,35 @@ private static String readFile(Path path) throws IOException {
...
+	private Map<String, Map<String, String>> parseLangEntries(Path root) throws IOException {
+		Path langFolderPath = root.resolve("lang");
+		Map<String, Map<String, String>> allLanguagesMap = new HashMap<>();
+
+		if (!Files.exists(langFolderPath)) {
+			return allLanguagesMap;
+		}
+			Files.walk(langFolderPath, 1).filter(path -> !Files.isDirectory(path)).forEach(path -> {
+
+				Map<String, String> currentLanguageMap = Maps.newHashMap();
```
coderbot16, line 110: One less import and actually fewer characters overall
```suggestion
				Map<String, String> currentLanguageMap = new HashMap<>();
```

coderbot16 (changes requested):
These are good features, but I would really prefer it if there were separate PRs for each (1 PR for the reloading, 1 PR for the translations) because it makes it much easier to get the good stuff merged quickly and focus on the stuff that needs fixing

OverlordsIII:
Please look at #25 for the translations. After that, I will add the reloading in a seperate pr

PR #30, opened 2020-12-27
Added the abillity to see the currently loaded shader in the options menu

hYdos:
Yes


coderbot16:
I don't think that Iris should clutter up the options screen like this - maybe the f3 screen is more appropriate?

hYdos:
> 
> 
> I don't think that Iris should clutter up the options screen like this - maybe the f3 screen is more appropriate?

ok. ill move that code into the f3 menu :tiny_potato:

ColdIce1605:
kek

IMS212:
made #33 to succeed this

coderbot16:
Since #33 was merged, I will close this for now.

However, having a proper Iris configuration gui (maybe accessed through the options menu?) would be pretty great, if you're interested in that.

PR #35, opened 2020-12-28
Add zip shaderpack support

OverlordsIII:
Adds support for zipped shaderpacks.


src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -91,6 +100,21 @@ private void loadExternalShaderpack(String name) {
 		logger.info("Using shaderpack: " + name);
 	}
 
+	private static Optional<Path> loadExternalZipShaderpack(Path shaderpackPath) {
+		try {
+			FileSystem zipSystem = FileSystems.newFileSystem(shaderpackPath, Iris.class.getClassLoader());
```
coderbot16, line 105: It's possible that this could cause a FileSystemAlreadyExistsException since you never close the created filesystem. This will cause issues with shaderpack reloading. It would be good to store this created filesystem so that it can be closed later. (such as when the shaderpack is reloaded)

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -91,6 +100,21 @@ private void loadExternalShaderpack(String name) {
...
+			Path root = zipSystem.getRootDirectories().iterator().next();//should only be one root directory for a zip shaderpack
+			return Files.walk(root)
+				.filter(Files::isDirectory)
+				.filter(path -> path.endsWith("shaders"))
+				.findFirst();
```
coderbot16, line 110: Why not just use [getPath](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/FileSystem.html#getPath(java.lang.String,java.lang.String...)) here?

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/Iris.java
coderbot16: This won't happen by itself but the reload PR could cause issues.

src/main/java/net/coderbot/iris/Iris.java
OverlordsIII: It's because while testing I found some zipped packs to have a folder before the `shaders` dir, so for example, `Trippy-Shaderpack-master.zip` could have a directory of `Trippy-Shaderpack-master.zip/Trippy-Shaderpack-master/shaders` so it's not always clear if the `shaders` dir is directly inside the zip or nested one folder

IMS212: Couldn’t this cause issues with chopapic, where it has multiple shaders in one zip?

src/main/java/net/coderbot/iris/Iris.java
OverlordsIII: Okay, ill just store it for now, and if this pr gets merged i'll fix it in the reloading pr.

src/main/java/net/coderbot/iris/Iris.java
coderbot16: Ah, right. Maybe it could be better to look for `shaders` dir, and if it doesn't exist, fall back to this method? Not sure though.

In any case, it would definitely help to have a comment explaining that in the code.

OverlordsIII: Okay I think i've added everything, can you take a final look around?

coderbot16 (approved):
Looks good, thanks!

PR #41, opened 2020-12-28
[Actions] Build any branch, cache Gradle packages and save artifacts

altrisi:
This PR adds the features in the title to the current Github Action.

It builds on every branch because there is no reason not to. It can also help checking if a PR builds correctly when targetting a different branch than `trunk`.

It uses the Actions `cache`, caching files produced by Gradle (dependencies, etc), so builds are faster.

And it saves artifacts in the action run, making them available to be downloaded from Github to check what could have gone wrong (if the user has a Github account). Those are not releases, and are only stored for 30 days IIRC.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,12 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
+    - name: Cache Gradle packages
+      uses: actions/cache@v2
+      with:
+        path: ~/.gradle/caches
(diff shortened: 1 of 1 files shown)


coderbot16 (approved):
Seems reasonable enough. I don't fully trust the cache to not get messed up somewhere along the way, but that would be simple enough to fix. Thanks!

altrisi:
Well, the cache is only applied if all `.gradle` files have the same hash, so in theory it gets reset when adding/removing dependencies, etc, so it shouldn't cause problems.

coderbot16:
Makes sense, sounds good.

PR #42, opened 2020-12-28
Make camera in fluid check use actual camera position

altrisi:
This started being a fix for camera having the lava effect when the player was just barely touching lava (first commit), but ended up changing the way the effect is decided.

This is needed because in third person, where the entity _eyes_ are submerged may not be where the _camera_ is submerded.

This is actually the way the game does it (in `net.minecraft.client.render.BackgroundRenderer#applyFog`), though I haven't checked if OF does the same (I'd suppose it does).

BTW, talking about fluids, Iris internal currently removes the vanilla effect of being submerged in fluids, not sure if that's known (although according to the debugger, that `applyFog` function is still called).


coderbot16 (approved):
Looks good to me, thanks!

I've observed the overlay being removed, but I haven't figured out exactly what causes it. I've created an issue for it: https://github.com/IrisShaders/Iris/issues/43

PR #47, opened 2020-12-29
Added langs

maxigator:
Added langs : 
de_de
en_gb
es_es
fr_fr
it_it
ja_jp
nl_nl
pr_pt
ru_ru

Let me know if you want me to change the way i translated

### DIFF
--- a/src/main/resources/assets/iris/lang/de_de.json
+++ b/src/main/resources/assets/iris/lang/de_de.json
@@ -1,0 +1,6 @@
+{
+  "iris.shaders.reloaded": "Shader aktualisiert!!",
+  "iris.keybind.reload": "Aktualisieren der Shader",
+  "iris.keybinds": "Iris",
+  "iris.shaders.reloaded.failure": "Nicht aktualisierte Shader: %s"
+}
--- a/src/main/resources/assets/iris/lang/de_de.json
+++ b/src/main/resources/assets/iris/lang/de_de.json
@@ -1,5 +1,5 @@
 {
-  "iris.shaders.reloaded": "Shader aktualisiert!!",
+  "iris.shaders.reloaded": "Shader aktualisiert!",
   "iris.keybind.reload": "Aktualisieren der Shader",
   "iris.keybinds": "Iris",
   "iris.shaders.reloaded.failure": "Nicht aktualisierte Shader: %s"
--- a/src/main/resources/assets/iris/lang/de_de.json
+++ b/src/main/resources/assets/iris/lang/de_de.json
@@ -1,6 +1,5 @@
 {
-  "iris.shaders.reloaded": "Shader aktualisiert!",
-  "iris.keybind.reload": "Aktualisieren der Shader",
-  "iris.keybinds": "Iris",
-  "iris.shaders.reloaded.failure": "Nicht aktualisierte Shader: %s"
+  "iris.shaders.reloaded": "Shader neu geladen!",
+  "iris.keybind.reload": "Shader neu laden",
+  "iris.shaders.reloaded.failure": "Shader neu laden fehlgeschlagen! Grund: %s"
 }
--- a/src/main/resources/assets/iris/lang/en_gb.json
+++ b/src/main/resources/assets/iris/lang/en_gb.json
@@ -1,6 +1,0 @@
-{
-  "iris.shaders.reloaded": "Reloaded Shaders!",
-  "iris.keybind.reload": "Reload Shaders",
-  "iris.keybinds": "Iris",
-  "iris.shaders.reloaded.failure": "Failed to Reload Shaders! Reason: %s"
-}
--- a/src/main/resources/assets/iris/lang/en_gb.json
+++ b/src/main/resources/assets/iris/lang/en_gb.json
@@ -1,0 +1,6 @@
+{
+  "iris.shaders.reloaded": "Reloaded Shaders!",
+  "iris.keybind.reload": "Reload Shaders",
+  "iris.keybinds": "Iris",
+  "iris.shaders.reloaded.failure": "Failed to Reload Shaders! Reason: %s"
+}
--- a/src/main/resources/assets/iris/lang/es_es.json
+++ b/src/main/resources/assets/iris/lang/es_es.json
@@ -1,0 +1,6 @@
+{
+  "iris.shaders.reloaded": "Shaders actualizados!",
+  "iris.keybind.reload": "Actualizar los Shaders",
+  "iris.keybinds": "Iris",
+  "iris.shaders.reloaded.failure": "Fallo en la actualización de los shaders: %s"
+}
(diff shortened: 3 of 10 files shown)


Yales21:
There is a little mistake with the French. It is not "Shaders actualisé**e**s !" but ”Shaders actualisés !"

maxigator:
Are you sure because I thought that "shaders" was feminine because its "ombres" which is feminine so idk

maxigator:
Oh sorry I just looked it up and its un shader and not une shader

Pepe20129:
The spanish translation is ok.

wurunduk:
Russian translation is quite wrong, here is a fix:
`
{
  "iris.shaders.reloaded": "Шейдеры перезагружены!",  
  "iris.keybind.reload": "Перезагрузить шейдеры",  
  "iris.keybinds": "Iris",  
  "iris.shaders.reloaded.failure": "Ошибка при перезагрузке шейдеров: %s"  
} 
`

src/main/resources/assets/iris/lang/es_es.json
```diff
@@ -0,0 +1,6 @@
+{
+  "iris.shaders.reloaded": "Shaders actualizados!",
+  "iris.keybind.reload": "Actualizar los Shaders",
+  "iris.keybinds": "Iris",
+  "iris.shaders.reloaded.failure": "Fallo en la actualización de los shaders: %s"
```
altrisi, line 5: I think the correct word in Spanish for "reload" would be "recargar", hence the following suggestion:
```suggestion
  "iris.shaders.reloaded": "Shaders recargados!",
  "iris.keybind.reload": "Recargar Shaders",
  "iris.keybinds": "Iris",
  "iris.shaders.reloaded.failure": "Error al recargar los shaders: %s"
```
I'd say "actualizar" is more like "update".

src/main/resources/assets/iris/lang/en_gb.json
```diff
@@ -0,0 +1,6 @@
+{
```
coderbot16, line 1: If you're not changing anything from the en_us file, then no need to have this to begin with.

src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -0,0 +1,6 @@
+{
+  "iris.shaders.reloaded": "Shader aktualisiert!",
+  "iris.keybind.reload": "Aktualisieren der Shader",
+  "iris.keybinds": "Iris",
```
coderbot16, line 4: Since you're not changing this, don't add it to the translation files because it will automatically fall back to en_us. It's only necessary to add entries for things that differ from en_us.
```suggestion
```

src/main/resources/assets/iris/lang/ru_ru.json
```diff
@@ -0,0 +1,6 @@
+{
+  "iris.shaders.reloaded": "Шейдеры перезагружены!",
+  "iris.keybind.reload": "Перезагрузить шейдеры",
+  "iris.keybinds": "Iris",
+  "iris.shaders.reloaded.failure": "Ошибка при перезагрузке шейдеров: %s"
```
coderbot16, line 5: JustMarkov translated this slightly differently at https://github.com/JustMarkov/Iris/commit/d8dd925bad0a2b521beeefdd1a5155ce52dc1e9f :

```
  "iris.shaders.reloaded.failure": "Ошибка при Перезагрузке шейдеров! Причина: %s"
```

Not sure which is better, however I think JustMarkov's version is closer in form to the original English version.

However, it's possible that "Failed to reload shaders: %s" like what you did might be a better way to write things in the english version to... I'm split on this.

coderbot16 (changes requested):
The translations seem reasonable enough, the only problem at the moment is that it looks like only fr_fr, es_es, and ru_ru have been verified though. I have no way to know whether the other translations are good or not, and I guess that just calls into question whether it's better to have translations that potentially have small errors or no translations at all. In any case, the translations seem to look fine to me otherwise.

Thanks for your contributions!

src/main/resources/assets/iris/lang/ru_ru.json
maxigator: I decided to change all error messages to the same as yours 
Like this "Failed to Reload Shaders! Reason: %s"

src/main/resources/assets/iris/lang/fr_fr.json
```diff
@@ -0,0 +1,5 @@
+{
+  "iris.shaders.reloaded": "Shaders Rechargés!",
+  "iris.keybind.reload": "Recharger Les Shaders",
+  "iris.shaders.reloaded.failure": "Echec lors du rechargement des shaders! Raison: %s"
```
arthurbambou, line 4: Upper case e of Echec needs an accent
```suggestion
  "iris.shaders.reloaded.failure": "Échec lors du rechargement des shaders! Raison: %s"
```

FoundationGames:
The GUI pull request is going to come in once the shader options pull request is merged, and the GUI PR has a lot of new lang entries that need to be translated... Either we can put the PR on hold or just add the new lang entries right now.
Here is the updated version of en_us.json in my fork (subject to change):
```json
{
  "iris.shaders.reloaded": "Reloaded Shaders!",
  "iris.keybind.reload": "Reload Shaders",
  "iris.keybinds": "Iris",
  "iris.shaders.reloaded.failure": "Failed to Reload Shaders! Reason: %s",

  "options.iris.apply": "Apply",
  "options.iris.openShaderPackFolder": "Open Shader Pack Folder...",
  "options.iris.refreshShaderPacks": "Refresh Shader Packs",
  "options.iris.shaderPackSelection": "Shader Packs...",
  "options.iris.shaderPackSelection.title": "Shader Packs",

  "option.iris.back": "Back",

  "pack.iris.select.title": "Select",
  "pack.iris.configure.title": "Configure",

  "page.iris.notFound": "Invalid Page!",
  "page.iris.noConfig": "No Shader Options.",

  "property.iris.boolean.true": "True",
  "property.iris.boolean.false": "False"
}
```

coderbot16:
I would rather just do these translations now, then add those ones in a future PR

arthurbambou (approved):

FoundationGames:
May be a bit inconvenient but I just added a few translations to my GUI, I updated my message above

maxigator:
I believe that `shaders reloaded` is better that `reloaded shaders`

src/main/resources/assets/iris/lang/fr_fr.json
```diff
@@ -0,0 +1,5 @@
+{
+  "iris.shaders.reloaded": "Shaders Rechargés!",
+  "iris.keybind.reload": "Recharger Les Shaders",
+  "iris.shaders.reloaded.failure": "Échec lors du rechargement des shaders! Raison: %s"
```
toinouH, line 4:
```suggestion
  "iris.shaders.reloaded.failure": "Échec lors du rechargement des shaders ! Raison: %s"
```
Same :)

toinouH (changes requested):
Almost ready :)

toinouH (approved):
Perfect :D

maxigator:
thanks for the help! in french I thought that there was a space before every "!".

coderbot16 (approved):
Looks good for now :+1:

Issue #51, opened 2020-12-30
Crash on AMD drivers while initializing shaders


coderbot16:
See this fix: https://github.com/grondag/canvas/commit/820bf754092ccaf8d0c169620c2ff575722d7d96

coderbot16:
Fixed in https://github.com/IrisShaders/Iris/commit/629fbed2a285a7879cfadb89a61d9b2bd364b43d

PR #58, opened 2021-01-01
Fix AMD driver issue (#51)

Shyrogan:
Fixes https://github.com/IrisShaders/Iris/issues/51.
This required using access wideners so I created one.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,6 +28,11 @@ dependencies {
 	includeFabricApiModule "fabric-api-base"
 }
 
+minecraft {
+	// Iris' access widener
+	accessWidener "src/main/resources/iris.accesswidener"
+}
+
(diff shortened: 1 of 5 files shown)


coderbot16 (commented):
Thanks! However, I do have a few issues with this approach:

1. It copies Grondag's code without any attribution / a license notice (required by the license that Canvas uses)
2. It messes with Minecraft shaders and forces us to copy decompiled code, I would prefer to just have our own shader abstractions

For the second point, I'm probably going to copy Sodium's GlShader / program code instead since it looks to be well made and be freely-licensed under the LGPL.

FoundationGames:
Is the access widener necessary? GlShader is a minecraft class so I'd think an invoker mixin would do fine

PR #62, opened 2021-01-02
Do not overwrite config file

dexman545:
Make IrisConfig#initialize do what it says in the JD.


coderbot16 (approved):
Looks good, thanks!

PR #65, opened 2021-01-02
Allow for completely turning off shaders and using vanilla rendering

FoundationGames:
The PR adds a builtin no-op shaderpack which is loaded when the shaderpack name is set to `(off)`.
The default shaderpack has also been changed to `(off)` rather than `(internal)`.

### DIFF
--- a/src/main/java/net/coderbot/iris/Iris.java
+++ b/src/main/java/net/coderbot/iris/Iris.java
@@ -72,7 +72,6 @@ public class Iris implements ClientModInitializer {
 
 		ClientTickEvents.END_CLIENT_TICK.register(minecraftClient -> {
 			if (reloadKeybind.wasPressed()){
-
 				try {
 					reload();
 					// TODO: Is this needed?
@@ -95,12 +94,14 @@ public class Iris implements ClientModInitializer {
 
 	public static void loadShaderpack() {
 		// Attempt to load an external shaderpack if it is available
-		if (!irisConfig.isInternal()) {
+		if (!irisConfig.isInternal() && !irisConfig.isNoOp()) {
 			if (!loadExternalShaderpack(irisConfig.getShaderPackName())) {
 				loadInternalShaderpack();
 			}
-		} else {
+		} else if(irisConfig.isInternal()) {
 			loadInternalShaderpack();
+		} else {
+			loadNoOpShaderPack();
 		}
 	}
 
@@ -172,6 +173,12 @@ public class Iris implements ClientModInitializer {
 		}
 
 		logger.info("Using internal shaders");
+	}
+
+	private static void loadNoOpShaderPack() {
+		currentPack = ShaderPack.NO_OP;
+
+		logger.info("Using no shaders");
 	}
 
 	public static void reload() throws IOException {
--- a/src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
+++ b/src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
@@ -14,6 +14,7 @@ import java.util.Properties;
 import net.coderbot.iris.Iris;
 import net.coderbot.iris.gl.texture.InternalTextureFormat;
 import org.apache.logging.log4j.Level;
+import org.jetbrains.annotations.Nullable;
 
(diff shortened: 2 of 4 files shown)


Pepe20129 (commented):
I think it looks good but I'm not sure.

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -241,4 +250,15 @@ public boolean isValid() {
 			}
 		}
 	}
+
+	static {
+		ShaderPack s;
+		try {
+			s = new ShaderPack(null);
+		} catch (IOException e) {
+			e.printStackTrace();
```
OverlordsIII, line 259: Maybe provide some more context via a call to Iris's logger and the specific error?

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -174,6 +175,12 @@ private static void loadInternalShaderpack() {
 		logger.info("Using internal shaders");
 	}
 
+	private static void loadNoOpShaderPack() {
+		currentPack = ShaderPack.NO_OP;
```
OverlordsIII, line 179: If there was a problem loading the `NO_OP` shaderpack, it get's set to null [here](https://github.com/FoundationGames/Iris/blob/disable_shaders/src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java#L260). Currently if the currentPack is also null, it will just throw at a `requireNull` check [here](https://github.com/IrisShaders/Iris/blob/trunk/src/main/java/net/coderbot/iris/Iris.java#L233), so maybe check if currentPack is null and then fallback to internal if it is with a log error?

OverlordsIII (changes requested):

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
FoundationGames: The try-catch is there simply because the constructor `throws IOException`. With the way I have things, an `IOException` should never actually be thrown from the `ShaderPack` constructor (and methods throwing `IOException` that are called in the constructor) when the passed Path is `null`. The try-catch is there due to the constructor's method signature already containing `throws IOException`.

FoundationGames: A likely change would be to just ignore the exception.

src/main/java/net/coderbot/iris/Iris.java
FoundationGames: Ideally, the no-op pack should never fail to load (see above), since it's hardcoded. One of the reasons to add the no-op shader pack is that it's a more reliable fallback than the internal shaders, and that it could become the new fallback (may be implemented in a future commit if this isn't a bad idea in some way).

OverlordsIII (approved):

coderbot16:
Hi,

Sorry about not reviewing this PR. However, I've decided that I want to approach this differently, based off of my work in the `render-layer-experiments` branch. Essentially, right now there is a `ShaderPipeline` class that handles setting up and tearing down the current shader state. I would like to split this up into a few classes:

* `FixedFunctionWorldPipeline`: Essentially a no-op pipeline (equivalent in result to `(OFF)` as implemented in this PR), doesn't use any shaders and should just match vanilla rendering
* `ForwardWorldPipeline`: For the few shaderpacks that don't use any deferred rendering techniques (deferred programs, composite programs, final programs, custom draw buffers, etc), this could allow the usage of fabulous graphics with extremely basic shaderpacks
* `DeferredWorldPipeline`: Closer to the current ShaderPipeline class, this would apply to most shaderpacks.

Notably, in the `(OFF)` case, no shaderpack would be loaded to begin with. I would also like to implement this as a separate "enable shaders" option instead of using the special-case `(off)` key. This could also allow toggling on and off shaders as well, and could be represented in the shaderpack GUI as a grayed-out shaderpack list with an enable / disable checkbox.

Also, I would like to work on fixing issues with internal shaders so that they match vanilla rendering, so that it isn't as critical to be able to turn shaders off.

PR #70, opened 2021-01-03
Fix #64

### LINKED ISSUE #64: Print a more useful error when loading a corrupted ZIP shaderpack
"zip END header not found" isn't very clear, it's more useful to just say that the zip is corrupted and to tell the user to download it again.

dexman545:
Untested, but should work. Informs the user that the shaderpack is corrupted and needs to be downlaoded again

Closes #64

### DIFF
--- a/src/main/java/net/coderbot/iris/Iris.java
+++ b/src/main/java/net/coderbot/iris/Iris.java
@@ -7,6 +7,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.zip.ZipException;
 
 import com.google.common.base.Throwables;
 import com.mojang.blaze3d.systems.RenderSystem;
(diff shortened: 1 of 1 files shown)


src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -154,7 +155,11 @@ private static boolean loadExternalShaderpack(String name) {
 				.filter(path -> path.endsWith("shaders"))
 				.findFirst();
 		} catch (IOException e) {
-			logger.error("Error while finding shaderpack for zip directory {}", shaderpackPath);
+			if (e instanceof ZipException && e.getMessage().equals("zip END header not found")) {
```
coderbot16, line 158: Why not just say the zip is corrupted on all instances of ZipException? There's a few other types of ZipException that can occur and they too seem to be caused by corrupted zip files. In this case, zip END header not found seems to indicate that the file was truncated.

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -154,7 +155,11 @@ private static boolean loadExternalShaderpack(String name) {
...
+				logger.error("The shaderpack appears to be corrupted, please try downloading it again {}", shaderpackPath);
```
coderbot16, line 159: The message is good but I wonder if it will be too hard for some people to spot. Maybe the shader pack GUI could have a better error dialog to go along with this?

coderbot16 (approved):
Thanks!

PR #91, opened 2021-01-28
Add a shaderpack selection GUI and Mod Menu config GUI

FoundationGames:
This PR adds a UI for selecting shader packs, accessible through Video Settings.
shader_selection
Shader pack configuration currently does not work, as Iris currently does not provide a framework for parsing commented `#define`s in shader programs.

The PR also introduces a configuration screen, accessible through Mod Menu or through the shader pack selection screen.
iris_config

Iris's config code has also been somewhat reworked to work with this configuration screen.


FoundationGames:
PR's not ready just yet, there's an issue with the UI being hidden after browsing the screens and resizing the window

src/main/java/net/coderbot/iris/gui/GuiUtil.java
```diff
@@ -0,0 +1,136 @@
...
+        tessellator.draw();
+    }
+
+    public static void drawCompactScrollBar(int x, int top, int bottom, int maxScroll, double scrollAmount, int maxPosition, float alpha) {
+        if (maxScroll > 0) {
+            int barHeight = (int)((float)((bottom - top) * (bottom - top)) / (float) maxPosition);
+            barHeight = MathHelper.clamp(barHeight, 32, bottom - top - 8);
+            int barTop = (int)scrollAmount * (bottom - top - barHeight) / maxScroll + top;
+            if (barTop < top) barTop = top;
+            int c = (((byte)(0x6E * alpha) << 24) | 0x0A0A0A);
```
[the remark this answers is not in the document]
FoundationGames: c is the color of the scroll bar background, d is the color of the scroll bar itself. notice that the scroll bar fades in when you hover over the list in game
I could probably name those two better, if that's necessary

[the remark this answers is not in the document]
TheGlitch76: When i was making PRs to coderbot's projects he was always pretty particular about naming and unlabled magic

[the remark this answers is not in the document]
FoundationGames: Alright, I can clean up anything unnamed/unclear
always helps in the end

src/main/java/net/coderbot/iris/config/IrisConfig.java
```diff
@@ -85,8 +158,77 @@ public void load() throws IOException {
...
+						Property[] ps = { ((PairProperty)p).getLeft(), ((PairProperty)p).getRight() };
+						for(Property q : ps) {
+							if(q instanceof ValueProperty) {
+								ValueProperty<?> vp = ((ValueProperty<?>)q);
+								vp.setValue(properties.getProperty(vp.getKey()));
+								vp.resetValueText();
+							}
+						}
+					}
+				}
```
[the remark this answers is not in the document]
FoundationGames: I think this is a good idea, I also hated that bit of code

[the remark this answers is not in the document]
FoundationGames: A method `forEvery()` is being added in the next commit, as well as a base class `TupleProperty` to allow for `Property`s with more values, and to make iterating through them easier

FoundationGames:
Don't consider merging this just yet, I'm in the middle of fixing bugs and concerns. Feel free to keep playtesting and reviewing what I have, though.

src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
```diff
@@ -0,0 +1,125 @@
...
+import net.coderbot.iris.Iris;
+import net.coderbot.iris.gui.GuiUtil;
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.font.TextRenderer;
+import net.minecraft.client.gui.widget.AlwaysSelectedEntryListWidget;
+import net.minecraft.client.util.math.MatrixStack;
+import net.minecraft.text.LiteralText;
+import net.minecraft.text.MutableText;
+import net.minecraft.util.Formatting;
+import net.minecraft.util.Tickable;
```
i509VCB, line 13: `TickableElement` is a thing of you are checking for an element which may tick.

src/main/java/net/coderbot/iris/gui/property/OptionProperty.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public T getValue() {
+        return values[index];
+    }
+
+    @Override
+    public void setValue(T value) {
+        List<T> vList = Lists.newArrayList(values);
```
i509VCB, line 40: By backing the implementation with a set or list I believe you can turn this right into a list.contains() -> do thing

src/main/java/net/coderbot/iris/gui/property/OptionProperty.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+        this.valueText = null;
+    }
+
+    protected boolean isButtonHovered(double mouseX, boolean entryHovered) {
+        return entryHovered && mouseX > cachedX + (cachedWidth * 0.6) - 7;
+    }
+
+    @Override
+    public boolean onClicked(double mouseX, double mouseY, int button) {
```
i509VCB, line 55: I kinda feel this all is just a lost button here, I'd assume value property set this up earlier, but it seems more appropriate to whip up a `ValuePropertyButton` rather than make the property value also a button.

src/main/java/net/coderbot/iris/gui/property/Property.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.text.Text;
+
+/**
+ * An element of a document. Used for options
+ * or configuration menus. Property objects are
+ * added to a PropertyList, and PropertyLists are
+ * added to a PropertyDocumentWidget which is used
+ * in a GUI.
+ */
+public class Property {
```
i509VCB, line 16: This seems like something that should be abstract?

src/main/java/net/coderbot/iris/shaderpack/ShaderProperties.java
```diff
@@ -32,9 +33,40 @@ public ShaderProperties(Properties properties) {
...
 		return new ShaderProperties();
 	}
+
+	public Properties asProperties() {
+		return properties;
+	}
+
+	private static class ImmutableProperties extends Properties {
+		@Override
+		public synchronized Object setProperty(String key, String value) {
```
i509VCB, line 49: I don't see why you need to synchronize methods that are `_ -> fail`

src/main/resources/fabric.mod.json
```diff
@@ -21,7 +21,10 @@
   "entrypoints": {
 	"client": [
 	  "net.coderbot.iris.Iris"
-	]
+	],
+    "modmenu": [
+      "net.coderbot.iris.config.IrisModMenu"
+    ]
```
i509VCB, line 27: There be indentation mismatches.

src/main/java/net/coderbot/iris/gui/property/OptionProperty.java
FoundationGames: OptionProperty is separate from ValueProperty. ValueProperty is a base class which is simply a property which stores a modifiable value. OptionProperty includes a UI for it all. I don't think a separate class for the button is necessary, however if anyone else has any input that would be appreciated.

src/main/java/net/coderbot/iris/gui/property/Property.java
FoundationGames: Sure. There aren't any abstract methods to inherit, but it makes more sense as it's a base class.

src/main/java/net/coderbot/iris/shaderpack/ShaderProperties.java
FoundationGames: I never noticed that it was implemented as a synchronized method... good catch, will be fixed

ColdIce1605:
@FoundationGames doesn't build

src/main/java/net/coderbot/iris/config/IrisConfig.java
```diff
@@ -6,12 +6,15 @@
 import java.util.*;
 
 import com.google.common.collect.ImmutableList;
+import javafx.beans.property.BooleanProperty;
```
ColdIce1605, line 9: bad

src/main/java/net/coderbot/iris/gui/property/Property.java
FoundationGames: I have decided not to make the class abstract since it may imply that it is not meant to be instantiated, which it is meant to be.

src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
FoundationGames: Fixed, will commit soon

src/main/resources/fabric.mod.json
FoundationGames: Fixed

src/main/java/net/coderbot/iris/config/IrisConfig.java
FoundationGames: will be running checkstyle from now on

src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
FoundationGames: Apparently extending `TickableElement` does not guarantee that child elements will be ticked...

src/main/java/net/coderbot/iris/gui/property/OptionProperty.java
FoundationGames: I don't think that kind of operation exists in lists/sets for `contains`, but I have decided to have the implementation rely on a list from the beginning rather than an array.

bluelhf:
Why not use `Collection#removeIf` when removing duplicate screens?

FoundationGames:
> Why not use Collection#removeIf when removing duplicate screens in ScreenStack.java?

Can't get back the instance with that method

coderbot16 (changes requested):
This PR is way too big for me to review.

I think that some serious consideration should be put into using an existing GUI library or similar, or perhaps making an ugly but minimal vanilla-like UI as a start. In any case, this PR really should probably be split up so that I can more feasibly review each part.

PR #97, opened 2021-02-01
Add some more simple uniforms

OverlordsIII:
Adds the following uniforms

- `playerMood`
- `heldBlockLightValue`
- `heldBlockLightValue2`
- `nightVision`
- `screenBrightness`

### DIFF
--- a/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
@@ -27,8 +27,6 @@ import net.minecraft.util.Hand;
 public final class CommonUniforms {
 	private static final MinecraftClient client = MinecraftClient.getInstance();
 
-	private static final Entity cameraEntity = client.getCameraEntity();
-
 	private CommonUniforms() {
 		// no construction allowed
 	}
@@ -57,6 +55,7 @@ public final class CommonUniforms {
 	}
 
 	private static float getBlindness() {
+		Entity cameraEntity = client.getCameraEntity();
 
 		if (cameraEntity instanceof LivingEntity) {
 			StatusEffectInstance blindness = ((LivingEntity) cameraEntity).getStatusEffect(StatusEffects.BLINDNESS);
@@ -79,6 +78,7 @@ public final class CommonUniforms {
 	}
 
 	private static float getNightVision() {
+		Entity cameraEntity = client.getCameraEntity();
 
 		if (cameraEntity instanceof LivingEntity) {
 
--- a/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
@@ -4,18 +4,24 @@ import static net.coderbot.iris.gl.uniform.UniformUpdateFrequency.ONCE;
 import static net.coderbot.iris.gl.uniform.UniformUpdateFrequency.PER_FRAME;
 
 import java.util.Objects;
+import java.util.function.IntSupplier;
 
 import net.coderbot.iris.gl.uniform.UniformHolder;
+import net.coderbot.iris.gl.uniform.UniformUpdateFrequency;
 import net.coderbot.iris.shaderpack.IdMap;
 import net.coderbot.iris.texunits.TextureUnit;
 
 import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.render.GameRenderer;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.effect.StatusEffectInstance;
 import net.minecraft.entity.effect.StatusEffects;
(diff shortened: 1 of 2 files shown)


src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -68,17 +70,39 @@ private static float getBlindness() {
...
 
+	private static float getPlayerMood() {
+		if (client.player == null) {
+			return 0.0F;
+		}
+		return client.player.getMoodPercentage();
+	}
+
 	private static float getNightVision() {
 		Entity cameraEntity = client.getCameraEntity();
```
Dragon-Seeker, line 81: Since blindness has the same variable call, possible to put it up top as:
`private static final Entity cameraEntity = client.getCameraEntity();`

Dragon-Seeker (changes requested):

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
coderbot16: That could cause weird issues because the camera entity can change over time (such as when you spectate a different entity in spectator mode),  so it shouldn't be treated like a constant. We can assume that the client is constant however.

OverlordsIII: Just added

OverlordsIII: oh oops, didn't get your message

OverlordsIII: I'll rollback

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -70,4 +102,31 @@ private static int isEyeInWater() {
...
+
+		private final Hand hand;
+
+		private HeldItemLightingSupplier(Hand targetHand) {
+			this.hand = targetHand;
+		}
+
+		@Override
+		public int getAsInt() {
+			if (client.player == null) {
```
coderbot16, line 116: This is fine as-is, but I do wonder whether it would be better to use the camera entity instead? It's an open question, since that would only have an effect in spectator mode in rare circumstances.

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -59,6 +70,27 @@ private static float getBlindness() {
 		return 0.0F;
 	}
 
+	private static float getPlayerMood() {
+		if (client.player == null) {
+			return 0.0F;
+		}
+		return client.player.getMoodPercentage();
```
coderbot16, line 77: Blank lines between braces and statements
```suggestion

		return client.player.getMoodPercentage();
```

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -59,6 +70,27 @@ private static float getBlindness() {
...
+
+		if (cameraEntity instanceof LivingEntity) {
+
+			LivingEntity livingEntity = (LivingEntity) cameraEntity;
+
+			if (livingEntity.getStatusEffect(StatusEffects.NIGHT_VISION) != null) {
+				return GameRenderer.getNightVisionStrength(livingEntity, CapturedRenderingState.INSTANCE.getTickDelta());
+			}
+		}
+		return 0.0F;
```
coderbot16, line 91: Ditto

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
OverlordsIII: For that uniform specifically, it probably wouldn't apply to spectators since it's for the item the player is holding, which in spectator should be nothing.

OverlordsIII: However for other uniforms, it probably should be kept as camera entity then so it gets updated for stuff like playerMood

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
OverlordsIII: I've added a new line but I don't think git has detected it, so I will just resolve this issue.

src/main/java/net/coderbot/iris/uniforms/ViewportUniforms.java
```diff
@@ -20,6 +20,8 @@
 	 */
 	private static final Window WINDOW = Objects.requireNonNull(MinecraftClient.getInstance().getWindow());
 
+	private static final MinecraftClient client = MinecraftClient.getInstance();
```
coderbot16, line 23: Dead code

coderbot16 (approved):
Thanks!

PR #107, opened 2021-02-12
Add shader config

OverlordsIII:
Adds a shader pack config for packs to add options that are toggled through a gui. The gui should get the different options via `ShaderPack#getConfig` and then calling `getBooleanOptions`, `getIntegerOptions`, `getFloatOptions`. Each `Option` contains multiple elements:
- comment
- default value
- value
- allowed values to set this value to (if empty you can set to any value)
- name

All Options are parsed via `ConfigOptionParser`.

Note: Options are not parsed from const variables _yet_

This is gonna be a draft because I think it probably needs a lot of reviews.


src/main/java/net/coderbot/iris/shaderpack/ConfigOptionParser.java
```diff
@@ -0,0 +1,299 @@
...
+	//Regex that matches for only integers and not floats
+	//Same as above but in the char class we remove the float specific checks
+		//remove matching a "."
+		//remove matching a "f" or a "F"
+	private static final Pattern INTEGER_OPTION_PATTERN = Pattern.compile("^(?<define>#define)\\s+(?<name>\\w+)\\s+(?<value>-?\\d+)\\s*(?<comment>(?<commentChar>//+)(?<commentContent>.*))?$");
+	//Some shaderpacks like sildurs have #define directives that are named with the program name
+	//like #define gbuffers_textured
+	//optifine does not use these in their config so we will not as well
+	//TODO figure what optifine does with these program defined names and implement
+	private static final Set<String> IGNORED_PROGRAM_NAMES = Util.make(new HashSet<>(), (set) -> {
```
FoundationGames, line 51: Would recommend you use multi-line comments throughout this whole thing rather than so many single line comments

src/main/java/net/coderbot/iris/shaderpack/ShaderPackConfig.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	public Map<String, Option<Float>> getFloatOptions() {
+		return floatOptions;
+	}
+
+	public Map<String, Option<Integer>> getIntegerOptions() {
+		return integerOptions;
+	}
+
```
FoundationGames, line 45: Considering that the fields are anyway final and that the getters don't do anything other than return each map, are getters really necessary?

FoundationGames (changes requested):

src/main/java/net/coderbot/iris/shaderpack/ConfigOptionParser.java
OverlordsIII: Yea sorry about that, I'll change them to multi-line comments

src/main/java/net/coderbot/iris/shaderpack/ShaderPackConfig.java
OverlordsIII: It's more of a personal preference, but It's not a big deal

OverlordsIII: Also, I've been thinking, there wasn't any need to even organize it as a map, since the key to value mapping is stored as the option's name to the option itself, which seems redundant as you won't need it anyway. I think it should probably be converted to a HashSet.

OverlordsIII: However, the values in the set/map shouldn't be modified either, so maybe add a getter that returns an immutable copy of the map, to make sure it's not tampered with

FoundationGames: I think if you want to do it that way, Immutable copies seems bit inefficient... What I'd recommend is have a method `Option<Integer> getIntegerOption(String name)` possibly with a method `Iterable<String> getIntegerOptions()` or something along those lines.

OverlordsIII: Okay I just finished adding this, so can you do a quick update on your branch?

FoundationGames: I think you may have misunderstood, there seems to no longer be a method for getting an option by key. I suggested adding two methods for each option type, one for getting by key and another for getting all available keys

OverlordsIII: Oh sorry about that, will fix now.

OverlordsIII:
I fixed a bunch of things and added some new functionality, so now I think this is ready for review:tm:

OverlordsIII:
I think i am going to close this for now because it's at 60 commits and only 15 of those are mine, so I will reopen it real quick

PR #166, opened 2021-03-08
Smooth "wetness" to match OF values

thedocruby:
Smoothed `wetness` uniform, choosing a value of `halfLife` so that the value decays 20 times during each 5-second transition from rainy to sunny weather. Removed associated TODO, as well as TODO completed in commit 67ed8f1c3404ca714dbad37526e3ea26e94430ee

### DIFF
--- a/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
@@ -63,7 +63,7 @@ public final class CommonUniforms {
 			.uniform1f(PER_FRAME, "screenBrightness", () -> client.options.gamma)
 			.uniform1f(PER_TICK, "playerMood", CommonUniforms::getPlayerMood)
 			.uniform2i(PER_FRAME, "eyeBrightness", CommonUniforms::getEyeBrightness)
-			.uniform2i(PER_FRAME, "eyeBrightnessSmooth", new SmoothedVec2f(10.0f, CommonUniforms::getEyeBrightness))
+			.uniform2i(PER_FRAME, "eyeBrightnessSmooth", new SmoothedVec2f(5.0f, CommonUniforms::getEyeBrightness))
 			.uniform1f(PER_TICK, "rainStrength", CommonUniforms::getRainStrength)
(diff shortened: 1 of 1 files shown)


ColdIce1605:
I don't know if this matches spec but this is the hl of wetness per optifine docs
/* WETNESSHL:600.0 */                           wetnessHalfLife = 600 (ticks)

thedocruby:
The whole process lasts 100 ticks, so that seems long to me, but i guess it makes sense. Thanks!

Also, where did you get this number? I would like to be able to find similar values in the future.

ColdIce1605:
https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt

thedocruby:
@ColdIce1605 Nice, Thanks! Very helpful document.

thedocruby:
I have updated `halfLife` values of `eyeBrightnessSmooth` and `wetness` to match OF values as defined [here](https://github.com/sp614x/optifine/blob/28172bc21b306334e06916d3e3907f251c51e0dc/OptiFineDoc/doc/shaders.txt#L490) and [here](https://github.com/sp614x/optifine/blob/28172bc21b306334e06916d3e3907f251c51e0dc/OptiFineDoc/doc/shaders.txt#L487), keeping in mind the Iris definition of `float halfLife`, as shown [here](https://github.com/IrisShaders/Iris/blob/67ed8f1c3404ca714dbad37526e3ea26e94430ee/src/main/java/net/coderbot/iris/uniforms/transforms/SmoothedFloat.java#L49).

coderbot16:
Actually, the Iris halfLife values are the same as the OptiFine halfLife values, but the OptiFine documentation is wrong. When it says 600 ticks, it actually means 600 tick-pairs like Iris (or 60 seconds). OptiFine's code matches the code of ShadersMod exactly here, and ShadersMod uses these "tick-pair" half life values.

It would be good to document this discrepancy in SmoothedFloat, however.

thedocruby:
@coderbot16 That's good to know! I have updated the values once again to account for the flawed documentation.

coderbot16:
Sounds good. And no need to ping me, I have notifications enabled for all issues and pull requests.

PR #169, opened 2021-03-08
Add StandardMacros

OverlordsIII:
Adds simple macros from [the optifine doc](https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L674-L716).


src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -5,12 +5,28 @@
 import net.coderbot.iris.gl.shader.ProgramCreator;
 import net.coderbot.iris.gl.shader.ShaderConstants;
 import net.coderbot.iris.gl.shader.ShaderType;
+import net.coderbot.iris.gl.shader.StandardMacros;
 import org.jetbrains.annotations.Nullable;
+import org.lwjgl.opengl.GL20C;
 import org.lwjgl.opengl.GL21C;
 
+import static net.coderbot.iris.gl.shader.StandardMacros.*;
```
coderbot16, line 13: `import static` should generally be avoided since it makes it unclear where things are being called from

src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -5,12 +5,28 @@
...
+
 public class ProgramBuilder extends ProgramUniforms.Builder {
 	private static final ShaderConstants EMPTY_CONSTANTS = ShaderConstants.builder().build();
 
+	private static final ShaderConstants MACRO_CONSTANTS = ShaderConstants.builder()
```
coderbot16, line 18: At some point this should be located elsewhere but this works for now

src/main/java/net/coderbot/iris/gl/shader/ShaderConstants.java
```diff
@@ -38,20 +36,36 @@ private Builder() {
 
         }
 
-        public void define(String name) {
-            this.define(name, EMPTY_VALUE);
+        public Builder define(String name) {
+        	if (name != null) {
```
coderbot16, line 40: Instead of silently ignoring null values, it's probably better to use Objects.requireNonNull

src/main/java/net/coderbot/iris/gl/shader/ShaderConstants.java
```diff
@@ -38,20 +36,36 @@ private Builder() {
...
-        public void define(String name, String value) {
+        public Builder define(String name, String value) {
             String prev = this.constants.get(name);
 
             if (prev != null) {
                 throw new IllegalArgumentException("Constant " + name + " is already defined with value " + prev);
             }
 
-            this.constants.put(name, value);
+            if (name != null && value != null) {
```
coderbot16, line 53: Same thing here, null values shouldn't silently be ignored

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,184 @@
...
+	/**
+	 * Gets the current mc version String in a 5 digit format
+	 *
+	 * @return mc version string
+	 * @see <a href="https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L677">Optifine Doc</a>
+	 */
+	public static String getMcVersion() {
+		String version = SharedConstants.getGameVersion().getReleaseTarget(); //release target so snapshots are set to the higher version
+
+		Matcher matcher = SEMVER_PATTERN.matcher(version);
```
coderbot16, line 33: A regex is a bit overkill for what is essentially a `string.split(".")`

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,184 @@
...
+
+		if (!matcher.matches()) {
+			return null;
```
coderbot16, line 36: Probably should not just silently return null... if we can't parse the game version, something is wrong!

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,184 @@
...
+	 * @see <a href="https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L683-L685">Optifine Doc for GLSL Version</a>
+	 */
+	public static String getGlVersion(int name) {
+		String info = GL20.glGetString(name);
+
+		Matcher matcher = SEMVER_PATTERN.matcher(Objects.requireNonNull(info));
+
+		if (!matcher.matches()) {
+			return null;
+		}
```
coderbot16, line 95: Similar comments here regarding the regex and silent failure

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,184 @@
...
+	 * This is done by calling {@link GL11#glGetString} with the arg {@link GL11#GL_EXTENSIONS}
+	 *
+	 * @see <a href="https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L713-L716">Optifine Doc</a>
+	 * @return list of activated extensions prefixed with "MC_"
+	 */
+	public static List<String> getGlExtensions() {
+		String[] extensions = Objects.requireNonNull(GL11.glGetString(GL11.GL_EXTENSIONS)).split("\\s+");
+
+		return Arrays.stream(extensions).map(s -> "MC_" + s).collect(Collectors.toList());
+	}
```
coderbot16, line 136: It doesn't seem like this detail is supported:

> Only the macros which are referenced and supported are added to the shader file.

While this is fine, that should probably be noted in the code somewhere.

coderbot16 (changes requested):
A few notes

It also looks like you squash merged the new changes from trunk, which messed up the git history a bit - it's not an issue for this PR, but please do not do this in the future

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
OverlordsIII: I am using regex because `glGetString` for gl version also returns extra unneeded info 
For example, my gl version is `4.6.0 NVIDIA 461.92`. However, the minecraft version one could most likely just be done like you said.

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,188 @@
...
+
+		String major = splitVersion[0];
+		String minor = splitVersion[1];
+		String bugfix = splitVersion[2];
+
+		if (bugfix == null) {
+			bugfix = "00";
+		}
+
+		if (minor == null || major == null) throw new IllegalStateException("Could not parse game version into a 5 digit format!");
```
coderbot16, line 47: These need to be changed to checks on the length of the split version, otherwise there will be `ArrayIndexOutOfBoundsException`s

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,188 @@
...
+
+		if (!matcher.matches()) {
+			throw new IllegalStateException("Could not parse GL version into 5 digit format");
+		}
+
+		String major = group(matcher, "major");
+		String minor = group(matcher, "minor");
+		String bugfix = group(matcher, "bugfix");
+
+		if (bugfix == null) { //if bugfix is not there, it is 0
```
coderbot16, line 102:
```suggestion
		if (bugfix == null) {
			// if bugfix is not there, it is 0
```

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,188 @@
...
+			bugfix = "0";
+		}
+
+		return major + minor + bugfix;
```
coderbot16, line 106: Maybe ensure that `major` and `minor` are non-null just to be certain

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,188 @@
...
+public class StandardMacros {
+
+	private static final Pattern SEMVER_PATTERN = Pattern.compile("(?<major>\\d+)\\.(?<minor>\\d+)\\.*(?<bugfix>\\d*)(.*)");
+
+
+	/**
+	 * Gets the current mc version String in a 5 digit format
+	 *
+	 * @return mc version string
+	 * @see <a href="https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L677">Optifine Doc</a>
```
coderbot16, line 28: These links are not pinned to a specific commit, prefer links of the form:

https://github.com/sp614x/optifine/blob/9c6a5b5326558ccc57c6490b66b3be3b2dc8cbef/OptiFineDoc/doc/shaders.txt#L696-L699

Otherwise, the line number references will gradually go out of date

coderbot16 (changes requested):
A few remaining comments, otherwise this should be good.

OverlordsIII:
I fixed those changes that you requested, could you check it again when you have time?

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,202 @@
...
+	public static String getMcVersion() {
+		String version = SharedConstants.getGameVersion().getReleaseTarget(); //release target so snapshots are set to the higher version
+
+		if (version == null) {
+			throw new IllegalStateException("Could not get the current minecraft version!");
+		}
+
+		String[] splitVersion = version.split("\\.");
+
+		if (splitVersion.length < 2) throw new IllegalStateException("Could not parse game version into a 5 digit format!");
```
coderbot16, line 39: Style: if statements should always use braces

Also print out the actual version for debugging purposes

```suggestion
		if (splitVersion.length < 2) {
			throw new IllegalStateException("Could not parse game version " + version + "!");
		}
```

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,202 @@
...
+
+
+	/**
+	 * Gets the current mc version String in a 5 digit format
+	 *
+	 * @return mc version string
+	 * @see <a href="https://github.com/sp614x/optifine/blob/9c6a5b5326558ccc57c6490b66b3be3b2dc8cbef/OptiFineDoc/doc/shaders.txt#L696-L699">Optifine Doc</a>
+	 */
+	public static String getMcVersion() {
+		String version = SharedConstants.getGameVersion().getReleaseTarget(); //release target so snapshots are set to the higher version
```
coderbot16, line 31: Style: comments on a separate line, with a space between the `//` and the comment text

Also clarified the implication of the comment

```suggestion
		// release target so snapshots are set to the higher version
		//
		// For example, 21w10a would return 1.17
		String version = SharedConstants.getGameVersion().getReleaseTarget();
```

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,202 @@
...
+	 */
+	public static String getOsString() {
+		switch (Util.getOperatingSystem()) {
+			case OSX:
+				return "MC_OS_MAC";
+			case LINUX:
+				return "MC_OS_LINUX";
+			case WINDOWS:
+				return "MC_OS_WINDOWS";
+			case SOLARIS: //Note: Optifine doesn't have a macro for Solaris. https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L689-L692
```
coderbot16, line 75: Should also use a fully qualified link here (with commit hash)
```suggestion
			case SOLARIS: // Note: Optifine doesn't have a macro for Solaris. https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L689-L692
```

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,202 @@
...
+		if (!matcher.matches()) {
+			throw new IllegalStateException("Could not parse GL version into 5 digit format");
+		}
+
+		String major = group(matcher, "major");
+		String minor = group(matcher, "minor");
+		String bugfix = group(matcher, "bugfix");
+
+		if (bugfix == null) {
+			//if bugfix is not there, it is 0
```
coderbot16, line 105:
```suggestion
			// if bugfix is not there, it is 0
```

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,202 @@
...
+			bugfix = "0";
+		}
+
+		if (major == null || minor == null) throw new IllegalStateException("Could not parse GL Version into 5 digit format!");
```
coderbot16, line 109: Similar thing to before
```suggestion
		if (major == null || minor == null) {
			throw new IllegalStateException("Could not parse GL version " + info + "!");
		}
```

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,202 @@
...
+	 * This is done by calling {@link GL11#glGetString} with the arg {@link GL11#GL_EXTENSIONS}
+	 *
+	 * @see <a href="https://github.com/sp614x/optifine/blob/9c6a5b5326558ccc57c6490b66b3be3b2dc8cbef/OptiFineDoc/doc/shaders.txt#L735-L738">Optifine Doc</a>
+	 * @return list of activated extensions prefixed with "MC_"
+	 */
+	public static List<String> getGlExtensions() {
+		String[] extensions = Objects.requireNonNull(GL11.glGetString(GL11.GL_EXTENSIONS)).split("\\s+");
+
+		//TODO note that we do not add extensions based on if the shader uses them and if they are supported
+		//see https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L738
```
coderbot16, line 142:
```suggestion
		// TODO note that we do not add extensions based on if the shader uses them and if they are supported
		// see https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.txt#L738
```

coderbot16 (changes requested):
Logic looks good, just a few remaining style issues.

coderbot16:
This will also need a small update to work with geometry shaders (replacement of a reference to EMPTY_CONSTANTS)

OverlordsIII:
Just fixed those style changes, could you check if everything's good?

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,209 @@
...
+	 * @see <a href="https://github.com/sp614x/optifine/blob/9c6a5b5326558ccc57c6490b66b3be3b2dc8cbef/OptiFineDoc/doc/shaders.txt#L705-L707">Optifine Doc for GLSL Version</a>
+	 */
+	public static String getGlVersion(int name) {
+		String info = GL20.glGetString(name);
+		System.out.println("GL Version: " + info);
+
+		Matcher matcher = SEMVER_PATTERN.matcher(Objects.requireNonNull(info));
+
+		if (!matcher.matches()) {
+			throw new IllegalStateException("Could not parse GL version into 5 digit format");
```
coderbot16, line 102: Note that it isn't a 5 digit format for the GL version, I think that it is fine to just state that the version couldn't be parsed, while including the correct version in the message, but without needing to state a detail that doesn't matter to the user (the fact that it needs to be parsed in order to create a fixed length version)

Same thing applies to the other places where versions are parsed.

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,209 @@
...
+		String minor = group(matcher, "minor");
+		String bugfix = group(matcher, "bugfix");
+
+		if (bugfix == null) {
+			// if bugfix is not there, it is 0
+			bugfix = "0";
+		}
+
+		if (major == null || minor == null) {
+			throw new IllegalStateException("Could not parse GL Version into 5 digit format!");
```
coderbot16, line 115: Ditto

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -0,0 +1,209 @@
...
+			// For example if we were running iris on 21w07a, getReleaseTarget() would return 1.17
+
+		if (version == null) {
+			throw new IllegalStateException("Could not get the current minecraft version!");
+		}
+
+		String[] splitVersion = version.split("\\.");
+
+		if (splitVersion.length < 2) {
+			throw new IllegalStateException("Could not parse game version \"" + version +  "\" into a 5 digit format!");
```
coderbot16, line 43: No need to state the usage of a 5 digit format here, that's just an internal detail.

coderbot16 (changes requested):

coderbot16 (approved):
Thanks!

PR #170, opened 2021-03-08
geometry shaders

ColdIce1605:
little broken fov affects it
its going to be marked as draft tell I figure it out


ColdIce1605:
Bug was do to incorrect inverse matrices fixed in latest iris commits.

src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -35,15 +36,22 @@ public static ProgramBuilder begin(String name, @Nullable String vertexSource, @
 			throw new RuntimeException("Failed to compile vertex shader for program " + name, e);
 		}
 
+		try {
+			geometry = new GlShader(ShaderType.GEOMETRY, name + ".gsh", geometrySource, EMPTY_CONSTANTS);
+		} catch (RuntimeException e) {
+			throw new RuntimeException("Failed to compile vertex shader for program " + name, e);
```
coderbot16, line 42:
```suggestion
			throw new RuntimeException("Failed to compile geometry shader for program " + name, e);
```

src/main/java/net/coderbot/iris/shaderpack/ProgramSource.java
```diff
@@ -38,7 +44,7 @@ public ProgramSet getParent() {
 	}
 
 	public boolean isValid() {
-		return vertexSource != null && fragmentSource != null;
+		return vertexSource != null && geometrySource != null && fragmentSource != null;
```
coderbot16, line 47: This logic is 100% broken. It's entirely valid for a program to lack a geometry shader, and most programs do in fact lack a geometry shader.

src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -35,15 +36,22 @@ public static ProgramBuilder begin(String name, @Nullable String vertexSource, @
 			throw new RuntimeException("Failed to compile vertex shader for program " + name, e);
 		}
 
+		try {
+			geometry = new GlShader(ShaderType.GEOMETRY, name + ".gsh", geometrySource, EMPTY_CONSTANTS);
```
coderbot16, line 40: This should NOT assume that the geometry shader is always present. Instead, the geometry shader should only be created if it is present, and later should only be attached if it is actually created.

coderbot16 (changes requested):
Geometry shader support is good, however it seems like this code was not tested with shaderpacks that don't use geometry shaders, since I see a few places where things will completely break if a pack does not use geometry shaders.

src/main/java/net/coderbot/iris/shaderpack/ProgramSource.java
ColdIce1605: what you're saying is it should be removed right?

coderbot16: yes, since it is valid for a shader program to lack a geometry shader

coderbot16 (commented):
A few stylistic issues but otherwise looks good

src/main/java/net/coderbot/iris/gl/shader/ProgramCreator.java
```diff
@@ -13,7 +13,7 @@ public static int create(String name, GlShader... shaders) {
 		int program = GL20C.glCreateProgram();
 
 		for (GlShader shader : shaders) {
-			GL20C.glAttachShader(program, shader.getHandle());
+			if(shader != null) GL20C.glAttachShader(program, shader.getHandle());
```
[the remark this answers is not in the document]
ColdIce1605: what is the wanted behavior then?

src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -23,27 +23,35 @@ public void bindAttributeLocation(int index, String name) {
...
 		GlShader vertex;
+		GlShader geometry;
 		GlShader fragment;
 
 		try {
 			vertex = new GlShader(ShaderType.VERTEX, name + ".vsh", vertexSource, EMPTY_CONSTANTS);
 		} catch (RuntimeException e) {
 			throw new RuntimeException("Failed to compile vertex shader for program " + name, e);
 		}
+	if(geometrySource != null)
```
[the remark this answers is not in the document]
ColdIce1605: `
if(geometrySource != null) {
		try {
			geometry = new GlShader(ShaderType.GEOMETRY, name + ".gsh", geometrySource, EMPTY_CONSTANTS);
		} catch (RuntimeException e) {
			throw new RuntimeException("Failed to compile geometry shader for program " + name, e);
		}
	} else {
		geometry = null;
	}
`
is this better?

src/main/java/net/coderbot/iris/gl/shader/ProgramCreator.java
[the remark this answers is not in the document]
coderbot16: If a null is passed to ProgramCreator, it should probably crash - nulls shouldn't be passed to ProgramCreator in the first place, imo

src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -35,15 +36,32 @@ public static ProgramBuilder begin(String name, @Nullable String vertexSource, @
...
+		int programId;
+		
+		if(geometry != null) {
+			programId = ProgramCreator.create(name, vertex, geometry, fragment);
+		} else {
+			programId = ProgramCreator.create(name, vertex, fragment);
+		}
 
 		vertex.destroy();
+		if (geometry != null) { geometry.destroy(); }
```
coderbot16, line 64:
```suggestion

		if (geometry != null) {
			geometry.destroy();
		}

```

src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -35,15 +36,32 @@ public static ProgramBuilder begin(String name, @Nullable String vertexSource, @
...
 		try {
 			fragment = new GlShader(ShaderType.FRAGMENT, name + ".fsh", fragmentSource, EMPTY_CONSTANTS);
 		} catch (RuntimeException e) {
 			throw new RuntimeException("Failed to compile fragment shader for program " + name, e);
 		}
 
-		int programId = ProgramCreator.create(name, vertex, fragment);
+		int programId;
+		
+		if(geometry != null) {
```
coderbot16, line 57:
```suggestion
		if (geometry != null) {
```

src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -35,15 +36,32 @@ public static ProgramBuilder begin(String name, @Nullable String vertexSource, @
 			throw new RuntimeException("Failed to compile vertex shader for program " + name, e);
 		}
 
+		if(geometrySource != null) {
```
coderbot16, line 39:
```suggestion
		if (geometrySource != null) {
```

coderbot16 (changes requested):
Looks good aside from a few minor remaining style issues

coderbot16 (approved):
Thanks!

PR #186, opened 2021-03-23
Properties conditional directive and macro parser in IdMap

Justsnoopy30:
Adds a parser for conditional directives (#if, #else, #endif) and macros (MC_VERSION) used for block.properties, item.properties, and entity.properties in IdMap.


coderbot16 (changes requested):
A few notes from an initial review:

* It looks like this includes an update of Gradle & Loom, these would preferably be kept to a separate PR
* It looks like this depends on #169, so I will need to review and approve that first before fully reviewing / approving this PR
* The preprocessor code should probably be moved to a separate file instead of being placed in IdMap
* Have you verified that nested conditions work with this approach?

coderbot16 (changes requested):
There are still a lot of changes to `gradlew` and `gradlew.bat` - could these be reverted please?

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -19,6 +17,9 @@
 import net.minecraft.util.Identifier;
 import net.minecraft.util.InvalidIdentifierException;
 
+import static java.lang.Integer.parseInt;
+import static net.coderbot.iris.shaderpack.PropertiesPreprocessor.parseProperties;
```
coderbot16, line 21: Avoid using `import static` where possible, prefer referencing those methods directly.

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -34,16 +35,16 @@
...
 
 	/**
-	 * a map that contains render layers for blocks in block.properties
+	 * A map that contains render layers for blocks in block.properties
 	 */
-	private Map<Identifier, RenderLayer> blockRenderLayerMap = Maps.newHashMap();
+	private Map<Identifier, RenderLayer> blockRenderLayerMap = new HashMap<>();
 
-	IdMap(Path shaderPath) {
+	public IdMap(Path shaderPath) {
```
coderbot16, line 47: Why does this need to be public?

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -59,20 +60,47 @@
 		// TODO: Properly override block render layers
 	}
 
+	/**
+	 * Loads properties from a properties file in a shaderpack path
+	 */
 	private static Optional<Properties> loadProperties(Path shaderPath, String name) {
-		Properties properties = new Properties();
+		if (shaderPath == null) return Optional.empty();
```
coderbot16, line 67: `if` statements should always be surrounded with braces. Also, why is this null check here in the first place? `shaderPath` should never be null.

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -59,20 +60,47 @@
...
+
+		String fileContents = readProperties(shaderPath, name);
+		if (fileContents == null) return Optional.empty();
```
coderbot16, line 70: Should have braces here too

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -99,7 +127,7 @@
 			int intId;
 
 			try {
-				intId = Integer.parseInt(key.substring(keyPrefix.length()));
+				intId = parseInt(key.substring(keyPrefix.length()));
```
coderbot16, line 130:
```suggestion
				intId = Integer.parseInt(key.substring(keyPrefix.length()));
```

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -59,20 +60,47 @@
...
 			return Optional.empty();
 		}
 
 		return Optional.of(properties);
 	}
 
+	private static String readProperties(Path shaderPath, String name) {
+		try {
+			return new String(Files.readAllBytes(shaderPath.resolve(name)), StandardCharsets.UTF_8);
+		} catch (IOException e) {
```
coderbot16, line 97: It's possible that this will be something other than a `NoSuchFileException`, that should be handled separately

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -59,20 +60,47 @@
...
+		String fileContents = readProperties(shaderPath, name);
+		if (fileContents == null) return Optional.empty();
 
+		List<String> lines = parseProperties(name, fileContents);
+
+		StringBuilder processed = new StringBuilder();
+		for (String line : lines) {
+			processed.append(line);
+			processed.append('\n');
+		}
```
coderbot16, line 78: Would probably be best to move this to PropertiesPreprocessor and have parseProperties return a string.

src/main/java/net/coderbot/iris/shaderpack/PropertiesPreprocessor.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import static net.coderbot.iris.gl.shader.StandardMacros.getMcVersion;
+
+public class PropertiesPreprocessor {
+	private static final ImmutableMap<String, String> MACRO_CONSTANTS = ImmutableMap.of("MC_VERSION", getMcVersion());
+	
+	public static List<String> parseProperties(String fileName, String fileContents) {
```
coderbot16, line 15: This should probably be named something like `process` or similar.

src/main/java/net/coderbot/iris/shaderpack/PropertiesPreprocessor.java
```diff
@@ -0,0 +1,140 @@
+package net.coderbot.iris.shaderpack;
+
+import com.google.common.collect.ImmutableMap;
+import net.coderbot.iris.Iris;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import static net.coderbot.iris.gl.shader.StandardMacros.getMcVersion;
```
coderbot16, line 10: Avoid using `import static` due to code clarity issues

src/main/java/net/coderbot/iris/shaderpack/PropertiesPreprocessor.java
```diff
@@ -0,0 +1,140 @@
...
+							break;
+						default: {
+							Iris.logger.error("Invalid operator " + operator + " in " + fileName);
+							continue;
+						}
+					}
+				} else {
+					Iris.logger.warn("Unknown variable fileName " + variable + " in " + fileName);
+					currentlyParsingConditionalProperties = false;
+				}
```
coderbot16, line 104: Should probably have a separate function for parsing the expression within the `#if` directive and returning whether or not it passed, then that would make it easier to support `#elif`.

src/main/java/net/coderbot/iris/shaderpack/PropertiesPreprocessor.java
```diff
@@ -0,0 +1,140 @@
...
+
+import static net.coderbot.iris.gl.shader.StandardMacros.getMcVersion;
+
+public class PropertiesPreprocessor {
+	private static final ImmutableMap<String, String> MACRO_CONSTANTS = ImmutableMap.of("MC_VERSION", getMcVersion());
+	
+	public static List<String> parseProperties(String fileName, String fileContents) {
+		List<String> lines = new ArrayList<>();
+
+		boolean currentlyParsingConditionalProperties = false;
```
coderbot16, line 18: Should have a more descriptive name for this... perhaps `currentConditionPassed` or something?

src/main/java/net/coderbot/iris/shaderpack/PropertiesPreprocessor.java
```diff
@@ -0,0 +1,140 @@
...
+		String currentConditional = null;
+
+		for (String line : fileContents.split("\\R")) {
+			String trimmedLine = line.trim();
+
+			if (trimmedLine.startsWith("#if ")) {
```
coderbot16, line 24: Since nested directives are not supported, that should cause a hard error somewhere for now instead of breaking silently

coderbot16 (changes requested):
A good start, noticed a few issues

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -34,16 +35,16 @@
...
 	 */
-	private Map<Identifier, Integer> blockPropertiesMap = Maps.newHashMap();
+	private Map<Identifier, Integer> blockPropertiesMap = new HashMap<>();
 
 	/**
-	 * a map that contains render layers for blocks in block.properties
+	 * A map that contains render layers for blocks in block.properties
 	 */
-	private Map<Identifier, RenderLayer> blockRenderLayerMap = Maps.newHashMap();
+	private Map<Identifier, RenderLayer> blockRenderLayerMap = new HashMap<>();
```
Kroppeb, line 45: Why these changes?

Justsnoopy30: Maps.newHashMap() is deprecated and should be replaced with new HashMap<>();

coderbot16:
I noticed that a few comments I made weren't fully addressed - I've resolved the ones that have been addressed, but there are a few that have not been addressed.

In particular, I'm not sure how `shaderPath` would end up being null, if this is the case then crashing is probably a better option.

Justsnoopy30:
> I noticed that a few comments I made weren't fully addressed - I've resolved the ones that have been addressed, but there are a few that have not been addressed, could you please review them?
> 
> In particular, I'm not sure how `shaderPath` would end up being null, if this is the case then crashing is probably a better option.

The shaderPath null check was a left-over from moving this from starline to trunk where completely disabling shaders (which can cause shaderPath to be null) isn't a thing. I'll remove that and address the other comments I missed tomorrow since I'm already in bed right now.

coderbot16:
Sounds good, thanks.

coderbot16 (approved):
Looks good, thanks! Now just waiting on a merge of the other PR.

Justsnoopy30:
> Looks good, thanks! Now just waiting on a merge of the other PR.

Do you still want me to create another method to parse the if statement or is it fine as-is?

coderbot16:
I'm willing to merge this PR as is, but that would definitely be nice as well.

coderbot16:
I squash merged the other PR, so it seems like this PR needs to be rebased on the current trunk, or the conflicts otherwise resolved.

coderbot16 (approved):
Thanks!

PR #187, opened 2021-03-23
Update to Gradle 7

Justsnoopy30:


ImperatorStorm:
i understand updating to loom 6, but why gradle 7?

DragonEggBedrockBreaking:
> i understand updating to loom 6, but why gradle 7? gradle 7 doesn't even have a stable release yet; latest is 6.8.3

Maybe JDK 16? Idk about gradlew inside files, but if you install gradle on your computer, and use that (with `gradle build` rather than `gradlew build`), then you need gradle 7 to build if you use JDK 16, so maybe that is the case for gradlew inside the source code?

ImperatorStorm:
> > i understand updating to loom 6, but why gradle 7? gradle 7 doesn't even have a stable release yet; latest is 6.8.3
> 
> Maybe JDK 16? Idk about gradlew inside files, but if you install gradle on your computer, and use that (with `gradle build` rather than `gradlew build`), then you need gradle 7 to build if you use JDK 16, so maybe that is the case for gradlew inside the source code?

why would you even be using java 16?

yitzy299:
Gradle 7 is still in beta, stay on 6.8.3, the latest

ImperatorStorm:
> Gradle 7 is still in beta, stay on 6.8.3, the latest

This

DragonEggBedrockBreaking:
> why would you even be using java 16?

Java 16 is the latest release now, and I read somewhere that older Java versions (except LTS versions ofc) stop getting more security updates, more patches, etc.

DragonEggBedrockBreaking:
> Gradle 7 is still in beta, stay on 6.8.3, the latest

I agree; If you're using Java 16, then you likely should just download Gradle 7 on your computer and build with that.

ImperatorStorm:
java 16 isn't even in arch repos, and neither is gradle 7

> older Java versions (except LTS versions ofc) stop getting more security updates, more patches, etc.

then why not use an LTS version?

Justsnoopy30:
Gradle 7 adds support for Java 16, without the gradle update, Java 16 fails to build the project.
Updating doesn't cause any problems, so there shouldn't be a problem with updating the wrapper.
In addition, Gradle 7 is a release candidate, not a prerelease, and as so it should be just as stable as the final release which will be released soon and this PR updated to use that.

yitzy299:
> Gradle 7 adds support for Java 16, without the gradle update, Java 16 fails to build the project.
> Updating doesn't cause any problems, so there shouldn't be a problem with updating the wrapper.
> In addition, Gradle 7 is a release candidate, not a prerelease, and as so it should be just as stable as the final release which will be released soon and this PR updated to use that.

1) Wait until a release, people can use the rc if they want
2) Why change part of the gradle files themselves? Just update the wrapper

ImperatorStorm:
>     2. Why change part of the gradle files themselves? Just update the wrapper

probably automatically done by gradle after attempting to build, or by IDE

DragonEggBedrockBreaking:
> Gradle 7 adds support for Java 16, without the gradle update, Java 16 fails to build the project.
> Updating doesn't cause any problems, so there shouldn't be a problem with updating the wrapper.
> In addition, Gradle 7 is a release candidate, not a prerelease, and as so it should be just as stable as the final release which will be released soon and this PR updated to use that.

As a JDK 16 user, I believe that we should just download Gradle 7 ourselves, and mods should only support it after it is released. You can download it here: https://services.gradle.org/distributions/ (`gradle-7.0-rc-1-bin.zip`) - remember to add `foldername/bin` to path :-)

DragonEggBedrockBreaking:
Update: this should now be changed to just updating Gradle, as my Loom 0.6 Pull Request got merged. This comment is, of course, assuming that Coderbot will keep this PR open.

PR #209, opened 2021-03-27
Add blockstate parsing

### LINKED ISSUE #32: Correctly handle entries in block.properties involving blockstates
OptiFine supports block states (instead of just blocks) in block.properties, we should too. This causes a lot of logspam with Sildur's and similar packs otherwise.

It's not 100% clear how to make this work seamlessly, consider the following examples:

```
block.0 = minecraft:stone_slab[half=top]
block.1 = minecraft:stone_slab[half=bottom]
```

```
block.0 = minecraft:stone_slab
```

We'll need to support both cases, because sometimes shaderpacks choose to differentiate based on state and sometimes they do not depending on the given block.

OverlordsIII:
Adds blockstate parsing for `IdMap` , should avoid those errors in the log.

Closes #32

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/vertices/MixinChunkRebuildTask.java
+++ b/src/main/java/net/coderbot/iris/mixin/vertices/MixinChunkRebuildTask.java
@@ -62,8 +62,7 @@ public class MixinChunkRebuildTask {
 			return -1;
 		}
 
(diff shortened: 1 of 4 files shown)


OverlordsIII:
Note: If this PR is merged, a small change has to be done to sodium to be able to update `mc_Entity` correctly. This requires first for trunk to be merged into the sodium compat branch and then a change can be added to sodium if needed.

PR #227, opened 2021-04-04
Add fogColor uniform

IMS212:
This will likely fix fog in the nether in multiple shaders, along with biome fog.


coderbot16:
This method works with vanilla but might mess up in the context of mods. https://github.com/CaffeineMC/sodium-fabric/blob/1.16.x/next/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/shader/ChunkFogMode.java is an example of a way that would work in all cases, though it is not 100% optimal because it could introduce pipeline stalls (which should be noted in a TODO) — as such, calls to that method should be minimized as much as possible. And, of course, in 1.17, it's possible to retrieve the fog mode directly from RenderSystem, which would make the issues with that previous approach sort of moot.

IMS212:
I've redone this PR to use a mixin into BackgroundRenderer instead of a manual calculation; this should be identical to vanilla's fog color now.

gradle.properties
```diff
@@ -3,7 +3,7 @@ org.gradle.jvmargs=-Xmx2G
 # Fabric Properties
 # check these on https://fabricmc.net/use
 minecraft_version=1.16.5
-yarn_mappings=1.16.5+build.5
+yarn_mappings=1.16.5+build.6
```
FoundationGames, line 6: Just wondering why this is necessary, but everything else looks great!

FoundationGames (approved):

gradle.properties
IMS212: Ah, that’s just a leftover from my Linux setup. For some reason build 5 gives an unexplainable error when trying to use runClient on my Arch setup, and build 6 doesn’t have any conflicts anyway so I don’t see any reason not to.

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -21,8 +21,10 @@
 import net.minecraft.item.BlockItem;
 import net.minecraft.item.ItemStack;
 import net.minecraft.tag.FluidTags;
+import net.minecraft.util.CubicSampler;
```
coderbot16, line 24: This seems unused.

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -21,8 +21,10 @@
...
 import net.minecraft.util.Hand;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.MathHelper;
```
coderbot16, line 27: Same here

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -32,6 +34,7 @@
 
 public final class CommonUniforms {
 	private static final MinecraftClient client = MinecraftClient.getInstance();
+	private static Vec3d fogColor;
```
coderbot16, line 37: This should probably be a member of CapturedRenderingState for the moment.

coderbot16 (changes requested):
This should be a good way to go about things until I figure out how to directly pull / save fog data from GlStateManager. I'd like a few things addressed first though.

coderbot16:
Apparently this causes a NullPointerException to occur on Complementary since the fog color isn't set early in the frame

IMS212:
All these issues should be solved now.

coderbot16 (approved):
Looks good to me, thanks!

PR #230, opened 2021-04-06
Cleanup

ColdIce1605:
Cleans up the ProgramBuilder class ie removes heavily duplicated code.

### DIFF
--- a/src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
+++ b/src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
@@ -44,15 +44,15 @@ public class ProgramBuilder extends ProgramUniforms.Builder {
 		GlShader geometry;
 		GlShader fragment;
 
-		vertex = createShaderType(ShaderType.VERTEX, name + ".vsh", vertexSource);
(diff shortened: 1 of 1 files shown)


ColdIce1605:
can I have a name check as im not sure if this is the right name for the method

OverlordsIII:
> can I have a name check as im not sure if this is the right name for the method

I think the method should probably just be called `createShader`?

src/main/java/net/coderbot/iris/gl/program/ProgramBuilder.java
```diff
@@ -88,4 +76,12 @@ public static ProgramBuilder begin(String name, @Nullable String vertexSource, @
 	public Program build() {
 		return new Program(program, super.buildUniforms());
 	}
+
+	private static GlShader createShaderType(ShaderType ShaderType, String name, @Nullable String source) {
```
OverlordsIII, line 80: the parameter `ShaderType` is uppercase and not camel case

ColdIce1605:
> 
> 
> > can I have a name check as im not sure if this is the right name for the method
> 
> I think the method should probably just be called `createShader`?

imma call it buildShader

PR #232, opened 2021-04-11
Gracefully handle packs not compiling

ColdIce1605:
Logger Compilation errors instead of crashing.
 * shouldn't work right now that why its a draft PR.


ColdIce1605:
Well I'm getting back to work on this. My current plan is to simplify this and implement the rest in later PRs.
1. switch to off when a error happens
2. Implement errors and a developer friendly system
3. settings to switch between the two.

ColdIce1605:
Well it turns out it default to off so the test was incorrect.

ColdIce1605:
Closed do to a newer commit will be opened back up afresh, for other things that need to be handled.
Turns out I was looking in the wrong area.
Oh well.

PR #238, opened 2021-04-12
Split UniformHolder interface

Kroppeb:
(Needed for Custom Uniforms)

This pr moves part of the code of `UniformHolder` into `LocationalUniformHolder` as the Custom Uniform input manager can't return any useful `location` nor can it accept arbitrary uniforms.

### DIFF
--- a/src/main/java/net/coderbot/iris/gl/uniform/Uniform.java
+++ b/src/main/java/net/coderbot/iris/gl/uniform/Uniform.java
@@ -3,7 +3,7 @@ package net.coderbot.iris.gl.uniform;
 public abstract class Uniform {
 	protected final int location;
 
-	public Uniform(int location) {
+	Uniform(int location) {
 		this.location = location;
 	}
 
--- a/src/main/java/net/coderbot/iris/gl/uniform/Uniform.java
+++ b/src/main/java/net/coderbot/iris/gl/uniform/Uniform.java
@@ -3,7 +3,7 @@ package net.coderbot.iris.gl.uniform;
 public abstract class Uniform {
 	protected final int location;
 
-	Uniform(int location) {
+	public Uniform(int location) {
 		this.location = location;
 	}
 
--- a/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
@@ -1,20 +1,15 @@
 package net.coderbot.iris.uniforms;
 
-import static net.coderbot.iris.gl.uniform.UniformUpdateFrequency.ONCE;
-import static net.coderbot.iris.gl.uniform.UniformUpdateFrequency.PER_FRAME;
-import static net.coderbot.iris.gl.uniform.UniformUpdateFrequency.PER_TICK;
-
 import java.util.Objects;
 import java.util.function.IntSupplier;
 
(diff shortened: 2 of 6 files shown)


src/main/java/net/coderbot/iris/uniforms/SamplerUniforms.java
```diff
@@ -32,7 +33,7 @@ private SamplerUniforms() {
 		// no construction allowed
 	}
 
-	public static void addCommonSamplerUniforms(UniformHolder uniforms) {
+	public static void addCommonSamplerUniforms(LocationalUniformHolder uniforms) {
```
coderbot16, line 36: Does this need to be a locational uniform holder, feels like UniformHolder work be acceptable here too?

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -31,14 +26,17 @@
...
+
 public final class CommonUniforms {
 	private static final MinecraftClient client = MinecraftClient.getInstance();
 
 	private CommonUniforms() {
 		// no construction allowed
 	}
 
-	public static void addCommonUniforms(UniformHolder uniforms, IdMap idMap) {
+	public static void addCommonUniforms(LocationalUniformHolder uniforms, IdMap idMap) {
```
coderbot16, line 39: Does this need to be locational?

src/main/java/net/coderbot/iris/gl/uniform/Uniform.java
```diff
@@ -3,7 +3,7 @@
 public abstract class Uniform {
 	protected final int location;
 
-	Uniform(int location) {
+	public Uniform(int location) {
```
coderbot16, line 6: Why does this need to be public?

coderbot16 (changes requested):
Looks mostly good, I have a few comments though. Also, this PR currently conflicts with trunk it appears.

src/main/java/net/coderbot/iris/uniforms/SamplerUniforms.java
Kroppeb: The samplers need the `location` to select the correct `shadow` sampler. It changes depending on whether the `watershadow` is used tr not it seems.

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
Kroppeb: yep, see https://github.com/IrisShaders/Iris/pull/238#discussion_r611900460

src/main/java/net/coderbot/iris/gl/uniform/Uniform.java
Kroppeb: Oh, this is shouldn't be in this commit. I think in one of my latest branches this doesn't even need to be public anymore. 

I'll remove it.

src/main/java/net/coderbot/iris/uniforms/SamplerUniforms.java
coderbot16: Right, forgot about that, in that case this is fine for now. Might be worth adding comments about that though.

coderbot16 (approved):
Thanks!

PR #247, opened 2021-04-18
Add basic build instructions

jasikpark:
I suppose that the idea is that you won't be pestered by people expecting support for compiling the mod without being paid for it, but I feel like it would be helpful for people who respect that concept but have no prior experience compiling java mods if there was a basic build instruction?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,9 @@ That being said... **If you want to test out compiled alpha builds of the mod, p
 
 **No support will be provided for users who cannot figure out how to properly compile the mod themselves.**
 
+Build:
+
+- `gradlew build`
 
 ## What is Iris, anyways?
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ That being said... **If you want to test out compiled alpha builds of the mod, p
 
 Build:
 
-- `gradlew build`
+- `chmod +x ./gradlew`
+- `./gradlew build`
 
 ## What is Iris, anyways?
 


DragonEggBedrockBreaking:
The whole point is that people *didn't* respect this idea, and that the support burden was too high and filled with new users reporting known issues in a not-very-understandable way...

jasikpark:
Well, I guess it would explain enough to let me 5 hours ago to get started easier 😅 I see your point though.

tedhong9:
Perhaps you could move the build instructions down to the very bottom and include a large disclaimer at the top of the section? I'm not sure if you could make something in markdown that lets you only click on hidden text only after checking that you've read the disclaimer, but that could also work. What do you all think?

DragonEggBedrockBreaking:
> Perhaps you could move the build instructions down to the very bottom and include a large disclaimer at the top of the section? I'm not sure if you could make something in markdown that lets you only click on hidden text only after checking that you've read the disclaimer, but that could also work. What do you all think?

This will still increase the report burden. Short of including some kind of timer, scroll speed counter, etc. there is probably no way to guarantee something has been read anyway. A disclaimer will anyway not help too much, as it is mostly inexperienced reporters reporting known bugs in a confusing way.

Also, I see many people trying to get "free information", and trying to help more people get shaders in sodium (while ignoring that Iris is WIP), yet I have yet to see these same people volunteering to help Coderbot deal with this burden, and agreeing that they will themselves any users who have arrived from their guides, mainly because although some people know how to compile a mod, only a fraction of them know how to deal with iris problems/support, and know iris' code, and can keep track of iris' issues.

squishycat92:
> Also, I see many people trying to get "free information", and trying to help more people get shaders in sodium (while ignoring that Iris is WIP), yet I have yet to see these same people volunteering to help Coderbot deal with this burden, and agreeing that they will themselves any users who have arrived from their guides, mainly because although some people know how to compile a mod, only a fraction of them know how to deal with iris problems/support, and know iris' code, and can keep track of iris' issues.
Yeah, I see your point, some people could just scroll past/click the checkbox without reading the disclaimer. In that case, it would probably be better to not include build instructions.

yitzy299:
> > Also, I see many people trying to get "free information", and trying to help more people get shaders in sodium (while ignoring that Iris is WIP), yet I have yet to see these same people volunteering to help Coderbot deal with this burden, and agreeing that they will themselves any users who have arrived from their guides, mainly because although some people know how to compile a mod, only a fraction of them know how to deal with iris problems/support, and know iris' code, and can keep track of iris' issues.
> 
> Yeah, I see your point, some people could just scroll past/click the checkbox without reading the disclaimer. In that case, it would probably be better to not include build instructions.

You clearly haven’t seen people trying to get support in the fabric discord: many of them will ping mods for things as simple as installing fabric API. I can assure you that there would be a large number of people trying to use release 0.1.0 of sodium with iris. I support this policy - if you know how to compile it, you should know how to fix issues with it

squishycat92:
> You clearly haven’t seen people trying to get support in the fabric discord: many of them will ping mods for things as simple as installing fabric API. I can assure you that there would be a large number of people trying to use release 0.1.0 of sodium with iris. I support this policy - if you know how to compile it, you should know how to fix issues with it

Wow, very interesting. I've always supported this policy (if you know how to compile it you can use it), but I didn't expect *that* much people to be clueless as to how mods work.

DragonEggBedrockBreaking:
> Wow, very interesting. I've always supported this policy (if you know how to compile it you can use it), but I didn't expect _that_ much people to be clueless as to how mods work.

Most people using iris just want shaders with sodium. Picture this: Somebody looks at a youtube video on how to get better fps, and they use optifine. They like shaders and sometimes use them. They then see another video on how Sodium is faster, so they download it, but then realist they can't use shaders. They just followed a tutorial on installing fabric + sodium, and just want some FPS and nice shaders. They then find out about iris from the CaffeineMC discord, and try to get it. Either it is easily accessible, and it doesn't work like expected, in which case they a) have no experience in reporting bugs, and b) don't realise how WIP it is. Or, it can be difficult to access. They leave it for a bit, come back when iris is released (in a few months? in a year? idk...). Imagine a hundred users like this. Now scale it even more.

coderbot16:
Most of what has been said here is accurate; adding compilation instructions wouldn't help developers and experienced users, the people who I currently want to be compiling the mod. Adding compilation instructions would also add quite a few mixed signals here, especially since it's placed directly after a line discouraging support for compiling the mod.

Iris just isn't ready for tons of people to be using it, and once it is, I'd prefer for people to use release builds (or snapshot builds from GH actions), not compiling the mod. As a result, I don't think that there will be a time where I want to strongly encourage compilation.

PR #250, opened 2021-04-28
Add a Shader Pack Selection UI

FoundationGames:
This PR adds a shader pack selection GUI.
The GUI allows for viewing available shader packs and selecting one, as well as applying the shader pack or cancelling the selection.
It also allows for previewing the shader pack's appearance with a transparent background.

shader_screen

As of opening, this PR is a draft.

### DIFF
--- a/src/main/java/net/coderbot/iris/Iris.java
+++ b/src/main/java/net/coderbot/iris/Iris.java
@@ -229,6 +229,20 @@ public class Iris implements ClientModInitializer {
 		internal = true;
 	}
 
+	public static boolean isValidShaderpack(Path pack) {
+		if (Files.isDirectory(pack)) {
+			return Files.exists(pack.resolve("shaders"));
+		}
+		if (pack.toString().endsWith(".zip")) {
+			try {
+				FileSystem zipSystem = FileSystems.newFileSystem(pack, Iris.class.getClassLoader());
+				return Files.exists(zipSystem.getPath("shaders"));
+			} catch (IOException ignored) {
+			}
+		}
+		return false;
+	}
+
 	public static void reload() throws IOException {
 		// allows shaderpacks to be changed at runtime
 		irisConfig.initialize();
--- a/src/main/java/net/coderbot/iris/Iris.java
+++ b/src/main/java/net/coderbot/iris/Iris.java
@@ -38,7 +38,7 @@ public class Iris implements ClientModInitializer {
 	public static final String MODID = "iris";
 	public static final Logger logger = LogManager.getLogger(MODID);
 
-	private static final Path shaderpacksDirectory = FabricLoader.getInstance().getGameDir().resolve("shaderpacks");
+	public static final Path shaderpacksDirectory = FabricLoader.getInstance().getGameDir().resolve("shaderpacks");
 
 	private static ShaderPack currentPack;
 	private static String currentPackName;
--- a/src/main/java/net/coderbot/iris/Iris.java
+++ b/src/main/java/net/coderbot/iris/Iris.java
@@ -2,7 +2,6 @@ package net.coderbot.iris;
 
 import java.io.IOException;
 import java.nio.file.*;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.zip.ZipException;
 
@@ -38,7 +37,7 @@ public class Iris implements ClientModInitializer {
 	public static final String MODID = "iris";
 	public static final Logger logger = LogManager.getLogger(MODID);
 
-	public static final Path shaderpacksDirectory = FabricLoader.getInstance().getGameDir().resolve("shaderpacks");
+	public static final Path SHADERPACKS_DIRECTORY = FabricLoader.getInstance().getGameDir().resolve("shaderpacks");
 
 	private static ShaderPack currentPack;
 	private static String currentPackName;
@@ -64,7 +63,7 @@ public class Iris implements ClientModInitializer {
 		);
 
 		try {
-			Files.createDirectories(shaderpacksDirectory);
+			Files.createDirectories(SHADERPACKS_DIRECTORY);
 		} catch (IOException e) {
 			Iris.logger.warn("Failed to create shaderpacks directory!");
 			Iris.logger.catching(Level.WARN, e);
@@ -131,7 +130,7 @@ public class Iris implements ClientModInitializer {
 	}
 
 	private static boolean loadExternalShaderpack(String name) {
-		Path shaderPackRoot = shaderpacksDirectory.resolve(name);
+		Path shaderPackRoot = SHADERPACKS_DIRECTORY.resolve(name);
 		Path shaderPackPath;
 
 		if (shaderPackRoot.toString().endsWith(".zip")) {
@@ -231,12 +230,20 @@ public class Iris implements ClientModInitializer {
 
 	public static boolean isValidShaderpack(Path pack) {
 		if (Files.isDirectory(pack)) {
-			return Files.exists(pack.resolve("shaders"));
+			try {
+				return Files.walk(pack)
+						.filter(Files::isDirectory)
+						.anyMatch(path -> path.endsWith("shaders"));
+			} catch (IOException ignored) {
+			}
 		}
 		if (pack.toString().endsWith(".zip")) {
 			try {
 				FileSystem zipSystem = FileSystems.newFileSystem(pack, Iris.class.getClassLoader());
-				return Files.exists(zipSystem.getPath("shaders"));
+				Path root = zipSystem.getRootDirectories().iterator().next();
+				return Files.walk(root)
+						.filter(Files::isDirectory)
+						.anyMatch(path -> path.endsWith("shaders"));
 			} catch (IOException ignored) {
 			}
 		}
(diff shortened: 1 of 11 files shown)


src/main/resources/assets/iris/lang/en_us.json
```diff
@@ -2,5 +2,14 @@
...
+  "iris.shaders.reloaded.failure": "Failed to Reload Shaders! Reason: %s",
+
+  "options.iris.apply": "Apply",
+  "options.iris.refresh": "Refresh",
+  "options.iris.openShaderPackFolder": "Open Shader Pack Folder...",
+  "options.iris.refreshShaderPacks": "Refresh Shader Packs",
+  "options.iris.shaderPackSelection": "Shader Packs...",
+  "options.iris.shaderPackSelection.title": "Shader Packs",
+
+  "pack.iris.select.title": "Select"
```
Motschen, line 14: I think it's a bit inconsistent to have some translation strings named "iris.something.something" and others "something.iris.something", so it'd be better to rename "options.iris.apply" to "iris.options.apply" for example.

FoundationGames: That would be inconsistent with much of Minecraft's naming scheme, for example: `item.minecraft.gold_ingot`
Minecraft options are simply keyed with `options.<option>` so I thought it would be cleaner and more consistent to keep the `option.` prefix. I'll need another opinion on this.

Motschen: Sorry for picking the wrong line, this was my first pr review

Motschen: > That would be inconsistent with much of Minecraft's naming scheme, for example: `item.minecraft.gold_ingot`
> Minecraft options are simply keyed with `options.<option>` 

Ah, makes sense. Minecraft's translation format is weird.

IMS212: It's highly likely `item.minecraft` exists to highlight the namespace as well, and not used globally. In my opinion mod ID should always go first.

IMS212: To extend this, I believe translations built off vanilla options should use vanilla's format, while others use the `modid.category.option` format

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -38,7 +38,7 @@
 	public static final String MODID = "iris";
 	public static final Logger logger = LogManager.getLogger(MODID);
 
-	private static final Path shaderpacksDirectory = FabricLoader.getInstance().getGameDir().resolve("shaderpacks");
+	public static final Path shaderpacksDirectory = FabricLoader.getInstance().getGameDir().resolve("shaderpacks");
```
OverlordsIII, line 41: should this be in snake case?

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -229,6 +229,20 @@ private static void loadInternalShaderpack() {
...
 	}
 
+	public static boolean isValidShaderpack(Path pack) {
+		if (Files.isDirectory(pack)) {
+			return Files.exists(pack.resolve("shaders"));
+		}
+		if (pack.toString().endsWith(".zip")) {
+			try {
+				FileSystem zipSystem = FileSystems.newFileSystem(pack, Iris.class.getClassLoader());
+				return Files.exists(zipSystem.getPath("shaders"));
```
OverlordsIII, line 239: Sometimes the shaders directory of the shaderpack is actually nested one directory, and Iris loads them irregardless, probably breaking the logic here.  See the comment [here](https://github.com/IrisShaders/Iris/pull/35#discussion_r549196794)

src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java
```diff
@@ -0,0 +1,147 @@
...
+		if (this.addedPackDialogTimer > 0) {
+			this.addedPackDialogTimer--;
+		}
+	}
+
+	@Override
+	public void filesDragged(List<Path> paths) {
+		List<Path> packs = paths.stream().filter(Iris::isValidShaderpack).collect(Collectors.toList());
+		for (Path pack : packs) {
+			if (Iris.isValidShaderpack(pack)) {
```
OverlordsIII, line 91: If the stream is already filtered using `isValidShaderpack`, won't this always be true?

src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java
```diff
@@ -0,0 +1,147 @@
...
+		if (!(base instanceof ShaderPackListWidget.ShaderPackEntry)) {
+			return;
+		}
+		ShaderPackListWidget.ShaderPackEntry entry = (ShaderPackListWidget.ShaderPackEntry)base;
+		String name = entry.getPackName();
+		Iris.getIrisConfig().setShaderPackName(name);
+		try {
+			Iris.reload();
+		} catch (IOException e) {
+			e.printStackTrace();
```
OverlordsIII, line 144: Maybe use `Iris.logger` to make it more clear when this happens?

src/main/java/net/coderbot/iris/mixin/MixinInGameHud.java
```diff
@@ -0,0 +1,26 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(InGameHud.class)
+public class MixinInGameHud {
+	@Shadow @Final private MinecraftClient client;
+
+	@Inject(method = "render", at = @At("HEAD"), cancellable = true)
+	public void handleTransparentScreens(MatrixStack matrices, float tickDelta, CallbackInfo ci) {
```
OverlordsIII, line 20: prefix with `iris$`?

OverlordsIII (changes requested):

src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java
FoundationGames: Forgot to remove that bit of code, thanks

src/main/java/net/coderbot/iris/Iris.java
FoundationGames: Correct. Although I do have a question - shouldn't the private static final fields also be in screaming snake case? Or have I misunderstood java conventions this whole time

src/main/java/net/coderbot/iris/mixin/MixinInGameHud.java
FoundationGames: Yep, thanks for reminding!

src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java
FoundationGames: Will do.

src/main/java/net/coderbot/iris/Iris.java
FoundationGames: You are correct, I will change the logic to match that of shader pack evaluation in the loading methods.

pengolod:
lgtm :p I've been using this and have had no problems as of yet..

altrisi:
I'd pass the exceptions to the logger as a second argument instead of making them print the stacktrace (to System.err).

Other than that and an `if` being too close to its condition looks good to me.

FoundationGames:
@altrisi Could you leave those comments on the file changes in a review so that I can see which lines exactly you'd prefer changed? I don't know which if statement you are talking about

altrisi:
I'll do that tomorrow, I'm on mobile sorry. It was in the Mixin to cancel rendering things when in the menu IIRC.

src/main/java/net/coderbot/iris/Iris.java
FoundationGames: Will mark conversation as resolved

FoundationGames:
@altrisi Your concerns were addressed in this latest commit, thanks for the feedback!

solonovamax:
You should perhaps look at adding a greyed out "Shader Options" button, that way it's there for when it is (eventually) implemented.

FoundationGames:
> You should perhaps look at adding a greyed out "Shader Options" button, that way it's there for when it is (eventually) implemented.

@solonovamax No, because that is not how the UI is going to be laid out in the end. Shader options will be a panel on the right. image

solonovamax:
Ah, okay. That seems quite nice tbh.
You'd probably want to add that box now (if you haven't already, which because you have a demo I assume you have ?), just so it's there and the UI won't need to be moved around in the future.

coderbot16:
You shouldn't add a box for shader options to the UI if it doesn't do anything. Just keep it to shaderpack selection for now, to avoid confusion.

MadDog443:
Just a suggestion about the shader UI, perhaps being able to select resource packs in the same menu would be a good quality of life improvement for many users? (IE where it says "Shader Packs" at the Very top make that a tab you can click on to switch between shader pack, resource pack, and/or data pack menus.)

FoundationGames:
> Just a suggestion about the shader UI, perhaps being able to select resource packs in the same menu would be a good quality of life improvement for many users? (IE where it says "Shader Packs" at the Very top make that a tab you can click on to switch between shader pack, resource pack, and/or data pack menus.)

@MadDog443 Absolutely not, the resource pack menu is separate from the shader pack menu just as is the data pack menu. And linking the menus will cause problems since the resource pack menu reloads resources (and pauses the game in the process) when you exit, meaning the experience between switching won't be very nice.
Overall the menus are unrelated and it is pointless to link any more menus to the shader pack select screen.

coderbot16:
Thanks for this! This GUI is a very nice addition to Iris and should make development and usage much more pleasant. Great work!

Issue #268, opened 2021-05-13
Shader pack selection UI includes the .minecraft/shaderpacks folder as an option if there is a non-zip shader pack installed


FoundationGames:
If you have a non-zip shaderpack installed, the shaderpack selection UI shows `shaderpacks` as an available option since `.minecraft/shaderpacks` contains a `shaders` folder in its file tree and was never checked for.

To reproduce:
- Add a folder to `.minecraft/shaderpacks` which contains a `shaders` folder, or add a `shaders` folder to `.minecraft/shaderpacks`
- Open shader pack selection UI

coderbot16:
Fixed in #269

Issue #263, opened 2021-05-09
No way to enable / disable shaders from the shader selection GUI


coderbot16:
Currently users must manually edit iris.properties to change whether shaders are enabled or not.

coderbot16:
Fixed in #269

PR #269, opened 2021-05-13
Add Enable/Disable Shaders GUI Option and Fix Bugs

### LINKED ISSUE #261: Cannot drag-and-drop directory shaderpacks into the shaderpack GUI
We need to do a recursive copy instead of just copying the folder itself, otherwise we end up with an empty shaderpack directory in the shaderpacks folder, which does not work.

FoundationGames:
Resolves #261, #268, #263
gui_fix_1


src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java
```diff
@@ -179,6 +181,20 @@ public void filesDragged(List<Path> paths) {
 		this.addedPackDialogTimer = 100;
 	}
 
+	private static void copyShaderPack(Path pack, String name) throws IOException {
+		Path target = Iris.SHADERPACKS_DIRECTORY.resolve(name);
+
+		// Make use of FileUtils to copy a directory with ease.
+		// Needs to convert the NIO Path to an IO File in order to work.
+		if (Files.isDirectory(pack)) {
+			FileUtils.copyDirectory(pack.toFile(), target.toFile());
```
coderbot16, line 190: Not all `Path`s can be converted to `File`s, this should be at the very least checked and noted in a comment.

src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
```diff
@@ -162,4 +186,38 @@ public void render(MatrixStack matrices, int index, int y, int x, int entryWidth
...
+		public EnableShadersButtonEntry(boolean enabled) {
+			this.enabled = enabled;
+		}
+
+		@Override
+		public void render(MatrixStack matrices, int index, int y, int x, int entryWidth, int entryHeight, int mouseX, int mouseY, boolean hovered, float tickDelta) {
+			GuiUtil.bindTexture(GuiUtil.IRIS_WIDGETS_TEX);
+			RenderSystem.color4f(1.0f, 1.0f, 1.0f, 1.0f);
+			RenderSystem.enableBlend();
+			RenderSystem.enableTexture();
```
coderbot16, line 202: If your goal is to abstract things, these need to be in GuiUtil too. Perhaps this should just be part of drawButton.

src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
```diff
@@ -30,30 +37,36 @@ public ShaderPackListWidget(MinecraftClient client, int width, int height, int t
...
-			int index = -1;
+			int index = 0;
 
 			for (String pack : BUILTIN_PACKS) {
 				index++;
 				addEntry(index, pack);
 			}
 
-			Collection<Path> folders = Files.walk(path, 1).filter(Iris::isValidShaderpack).collect(Collectors.toList());
+			Collection<Path> folders = Files.walk(path, 1).filter(ShaderPackListWidget::isValidPackPath).collect(Collectors.toList());
```
coderbot16, line 69: Perhaps try https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#list-java.nio.file.Path- instead? Seems like it would solve your problem in a nicer way:

> Some file systems maintain special links to the directory itself and the directory's parent directory. Entries representing these links are not included.

This is also probably relevant in the code for locating the `shaders` directory within a shaderpack.

src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
```diff
@@ -30,30 +37,36 @@ public ShaderPackListWidget(MinecraftClient client, int width, int height, int t
...
 	}
 
 	@Override
 	protected int getRowTop(int index) {
 		return super.getRowTop(index) + 2;
 	}
 
+	private static boolean isValidPackPath(Path path) {
+		return !path.equals(Iris.SHADERPACKS_DIRECTORY) && Iris.isValidShaderpack(path);
+	}
```
coderbot16, line 53: This should be removed - see my comment on the lines below

src/main/java/net/coderbot/iris/gui/GuiUtil.java
```diff
@@ -0,0 +1,83 @@
...
+ * Helps allow for easier portability
+ * to Minecraft 1.17 by abstracting
+ * some code that will be changed.
+ */
+public final class GuiUtil {
+	/**
+	 * Resource ID for the texture used for
+	 * Iris's GUI widgets.
+	 */
+	public static final Identifier IRIS_WIDGETS_TEX = new Identifier("iris", "textures/gui/widgets.png");
```
coderbot16, line 24: This should probably just be private, and you can then have a bindWidgetsTexture function in place of bindTexture.

src/main/java/net/coderbot/iris/gui/GuiUtil.java
```diff
@@ -0,0 +1,83 @@
...
+		return MinecraftClient.getInstance();
+	}
+
+	/**
+	 * Binds the texture to be used for
+	 * succeeding draw calls.
+	 *
+	 * @param texture The resource ID of the texture to be bound.
+	 */
+	public static void bindTexture(Identifier texture) {
```
coderbot16, line 38: This should probably be included within drawButton.

coderbot16 (changes requested):
Good improvements, have a few comments though.

src/main/java/net/coderbot/iris/gui/GuiUtil.java
FoundationGames: There's a specific reason why i didn't call to `bindTexture` directly in `drawButton`, and that is because `drawButton` can also draw a button with the vanilla widgets texture bound to appear like a vanilla button.

src/main/java/net/coderbot/iris/gui/GuiUtil.java
FoundationGames: Noted, however see my next few responses

src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
FoundationGames: Will be taken care of.

src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
FoundationGames: Will do.

src/main/java/net/coderbot/iris/gui/screen/ShaderPackScreen.java
FoundationGames: Alright, and I feel like there could be a better, cleaner solution to copying a directory and its contents, so if you know of one let me know

coderbot16: I've found a few examples that make use of `Files.walk`, but obviously it wouldn't be a one liner unless we have our own util class. I'm not seeing any one liners in the standard library but maybe I'm not looking hard enough. I wouldn't be opposed to that though, since relying on toFile isn't great, especially when we can be passed arbitrary `Path`s from the OS.

src/main/java/net/coderbot/iris/gui/GuiUtil.java
coderbot16: Makes sense. I think this is fine as long as the bindWidgetsTexture change is done.

FoundationGames: Also I believe this method is going to change in 1.17, and as something used very frequently in GUI code it would be easier to abstract it all so it can be changed quickly when porting

FoundationGames:
Converted to draft due to the discovery of a bug using `Files#list` - `list` does not go further than a depth of 1 when going through files in a directory

FoundationGames:
Bug has been fixed, PR is ready for review once again

coderbot16 (approved):
Thanks!

PR #307, opened 2021-06-01
Add a warning when the current shader pack uses unimplemented features

FoundationGames:
image

During loading, a shader pack can be marked as "using unimplemented features" and as such a warning will show.

As of the opening, this PR is a draft and the warning will be shown for any shader pack, there is no detection for unimplemented feature usage.


coderbot16:
Thanks for your contributions. I've decided to close this PR for the moment. When this PR was created, Iris was a little more than a week away from its initial release, with many shaders still unsupported. At the time, I requested this PR as part of an idea to try to warn the user about unsupported shaders, so that we could still reasonably release such an incomplete mod.

However, as it turns out, it wasn't really possible for these warnings to help a ton. Generally it's pretty obvious when Iris doesn't properly support a shader pack, because the pack will likely black screen or otherwise fail catastrophically. Also, detecting these unimplemented features is nontrivial and quite a project by itself, especially when it is also possible to spend time just implementing those features.

At this point, Iris isn't particularly far off from having initial implementations of all features of the OF shader pipeline. Once custom uniforms, compute shaders, proper image load store, and the things in shaders.properties are implemented, there won't be a ton of things remaining to implement. Once those are done, the remaining issues will be bugs and oversights, which can't easily be quantified in terms of a warning sign.

Hopefully, in the next few months, we'll be able to implement many of the remaining features & fix many of the bugs, but in the meantime, I'm not sure if fleshing out a warning screen that many users will ignore will be the best use of our time.

PR #313, opened 2021-06-03
Simplified Chinese translation

klkq:
Translation for Simplified Chinese. 
The translation for _shader pack_ is kept identical to that in Optifine.


coderbot16 (approved):
Thanks!

PR #318, opened 2021-06-06
new banner

LovelyA72:


coderbot16:
Banner looks nice, but maybe the bottom text could say "An open-source shaders mod compatible with OptiFine shaderpacks" so that it doesn't seem like the mod itself is compatible with OptiFine?

ColdIce1605:
Seems a little long but I couldn't come up with a better one.

coderbot16 (approved):
Thanks!

PR #327, opened 2021-06-08
Add keybinds to open the shaderpack selection screen and toggle shaders

Justsnoopy30:
This cherry-picks the extra keybinds to open the shaderpack screen and toggle shaders on and off from starline.


DragonEggBedrockBreaking (approved):
These keybinds are very useful, especially for those of us who use sodium and have to go through an extra step to go to the shader selection menu.

coderbot16 (approved):
Looks good to me. The Sodium fork will have a proper shaders button in the settings menu but these are helpful nonetheless. Thanks!

Issue #334, opened 2021-06-09
Water reflections and other features are broken at extreme distances


CalXee:
![2021-06-09_10 10 45](https://user-images.githubusercontent.com/68349758/121371590-a7aeb500-c90b-11eb-9e69-8b5dc93321bd.png)
![2021-06-09_10 10 35](https://user-images.githubusercontent.com/68349758/121371593-a8474b80-c90b-11eb-9811-f9ecb7ccb45c.png)

EminGT:
The reason is the cameraPosition uniform having floating point precission issues

coderbot16:
OptiFine keeps the cameraPosition uniform within a certain value range and simply ensures that `cameraPosition - previousCameraPosition` remains the same, since otherwise they are arbitrary. Whenever it leaves the range, cameraPosition and previousCameraPosition are both shifted by a fixed amount to keep cameraPosition within range.

PR #337, opened 2021-06-09
Update README.md to accommodate for #334

CalXee:


coderbot16:
I don't really think that many people will really encounter this? I just intended to call out some common potential bugs in the README. The issue tracker is best for everything else.

PR #342, opened 2021-06-09
Prevent enabling shaders in the GUI when no packs are present

FoundationGames:
You can't click the Enable/Disable Shaders button when there are no shader packs, and the button is forced to false during this state

(It also makes use of what was previously an unused button texture in Iris's assets, the disabled button texture)

### DIFF
--- a/src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
+++ b/src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
@@ -1,7 +1,6 @@
 package net.coderbot.iris.gui.element;
 
 import com.google.common.collect.ImmutableList;
-import com.mojang.blaze3d.systems.RenderSystem;
 import net.coderbot.iris.Iris;
 import net.coderbot.iris.gui.GuiUtil;
 import net.minecraft.client.MinecraftClient;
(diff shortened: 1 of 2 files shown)


src/main/java/net/coderbot/iris/gui/element/ShaderPackListWidget.java
```diff
@@ -86,6 +89,7 @@ public void addEntry(int index, String name) {
 		if (Iris.getIrisConfig().getShaderPackName().equals(name)) {
 			this.setSelected(entry);
 		}
+		packCount++;
```
coderbot16, line 92: packCount is never reset back to zero when the underlying list is cleared. If it's also possible for the underlying list to be modified in other ways, there are other ways for it to fall out of sync.

coderbot16 (changes requested):
Honestly after some thought I don't really like the behavior being implemented here. While it's weird for the button to say "Shaders Enabled" when there's no possible way for shaders to be enabled (ie, the user has no packs), I feel like it's worse to make it possible that the user dragging in a shader pack will forget to click the button to enable shaders as well, which could become a not uncommon support topic.

I'm not sure what the best solution to that is. Maybe we can keep shadersEnabled as true in the underlying settings, but just make the button say that shaders disabled if there are no packs. Then, the moment the user adds in a pack, it will say Shaders Enabled, revealing its true state.

coderbot16:
I've chosen to address this in a simpler way instead.

PR #354, opened 2021-06-10
Allow using fabulous when shaders are not enabled

### LINKED ISSUE #216: Cannot use Fabulous graphics option even when using vanilla rendering
With vanilla rendering enabled (shaderpack set to "null") I still cannot turn on Fabulous mode. There should be no technical limitations in turning on fabulous mode in this scenario right?

IMS212:
This implements the same patch used in starline, to allow using fabulous graphics when there are no shaders enabled.

Closes #216

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/fabulous/MixinOption.java
+++ b/src/main/java/net/coderbot/iris/mixin/fabulous/MixinOption.java
@@ -1,5 +1,6 @@
 package net.coderbot.iris.mixin.fabulous;
 
+import com.mojang.blaze3d.platform.GlStateManager;
 import net.coderbot.iris.Iris;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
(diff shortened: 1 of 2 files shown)


src/main/java/net/coderbot/iris/mixin/fabulous/MixinOption.java
```diff
@@ -15,6 +16,6 @@
 	@Redirect(method = "method_18554", at = @At(value = "INVOKE", target = "Lcom/mojang/blaze3d/platform/GlStateManager;supportsGl30()Z"))
 	private static boolean iris$onAttemptedToSelectFabulousGraphics() {
 		// Returning false here will cause Minecraft to cycle between Fancy and Fast, disabling Fabulous graphics
-		return false;
+		return !Iris.getIrisConfig().areShadersEnabled();
```
coderbot16, line 19: Instead of unconditionally returning true when shaders are enabled, we should call `supportsGl30()` first.

coderbot16 (changes requested):
Looks good, one change needed though.

src/main/java/net/coderbot/iris/mixin/fabulous/MixinOption.java
IMS212: Should be good now.

coderbot16 (approved):
Thanks!

PR #371, opened 2021-06-11
Add issue templates

NoComment1105:
Since Iris is getting bigger now and more Issues are coming in. A template set probably is viable for Iris to avoid unhelpful reports of issues. 
This contains templates for: Bug Reports, Crash Reports and Feature requests. It also contains a config.yml file to disable blank issues.


coderbot16 (approved):
Looks good to me.

PR #387, opened 2021-06-14
Fix cleared color on shadows

IMS212:
This fixes SEUS PTGI's raytracing, provided you do not use sodium.

### DIFF
--- a/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
+++ b/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
@@ -292,7 +292,7 @@ public class ShadowRenderer implements ShadowMapRenderer {
 
 		// TODO: Support shadow clear color directives & disable buffer clearing
 		// Ensure that the color and depth values are cleared appropriately
-		RenderSystem.clearColor(0.0f, 0.0f, 0.0f, 0.0f);
+		RenderSystem.clearColor(255, 255, 255, 1);
 		RenderSystem.clearDepth(1.0f);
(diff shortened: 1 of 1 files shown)


ColdIce1605:
My only question is this default behavior in OF, or shadermod because if it's not maybe this shouldn't be merged and the config option should be worked on which will take time of course.

JONTTM:
> My only question is this default behavior in OF, or shadermod because if it's not maybe this shouldn't be merged and the config option should be worked on which will take time of course.

SEUS PTGI works in OF by default

ColdIce1605:
No not what I meant, what I meant is this clear color the default in OF or does PTGI change it.

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -292,7 +292,7 @@ public void renderShadows(WorldRendererAccessor worldRenderer, Camera playerCame
 
 		// TODO: Support shadow clear color directives & disable buffer clearing
 		// Ensure that the color and depth values are cleared appropriately
-		RenderSystem.clearColor(0.0f, 0.0f, 0.0f, 0.0f);
+		RenderSystem.clearColor(255, 255, 255, 1);
```
coderbot16, line 295: Color values should be in the range of 0.0F to 1.0F when calling clearColor, not 0 to 255.

coderbot16 (changes requested):
The call to clearColor is slightly wrong. But I can confirm that the shadow color should be cleared to white, unless otherwise specified by the shader pack.

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
IMS212: Fixed.

PR #389, opened 2021-06-14
Fix end portal, block entity, and end sky shaders on 1.17

IMS212:


src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
```diff
@@ -232,4 +232,24 @@
...
+
+	@Inject(method = RENDER, at = @At(value = "CONSTANT", args = "stringValue=destroyProgress"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private void iris$EndBlockEntities(MatrixStack matrices, float tickDelta, long limitTime,
+									   boolean renderBlockOutline, Camera camera, GameRenderer gameRenderer,
+									   LightmapTextureManager lightmapTextureManager, Matrix4f matrix4f,
+									   CallbackInfo ci, Profiler profiler, Vec3d vec3d, double d, double e, double f,
+									   Matrix4f matrix4f2, boolean bl, Frustum frustum2, boolean bl3,
+									   VertexConsumerProvider.Immediate immediate) {
+		pipeline.setPhase(WorldRenderingPhase.OTHER);
+	}
```
coderbot16, line 254: I don't think that local capture is needed here. Also, the convention is `iris$startBlockEntities` afaik.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
IMS212: Fixed, would you mind reviewing again

coderbot16 (changes requested):
It looks like you didn't remove the stuff after the callbackinfo, which would probably cause the mixin to fail and the game to crash?

IMS212:
Fixed

coderbot16:
Thanks!

PR #390, opened 2021-06-14
Changes RenderTarget behavior to match OF per Niemand

ColdIce1605:
Should fix the issue posted in #development on the Discord.


coderbot16 (approved):
Thanks!

PR #391, opened 2021-06-15
Update README.md with a fix for some shaders failing to compile on Linux.

satyrnsstuff:
Fixed SEUS Renewed for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,8 @@ Iris has been progressing quite rapidly recently. The following shader packs mos
     * Underwater is broken
     * Weather is broken
 
-For Linux users running AMD GPUs, some shader packs can be fixed by launching the game with
-`allow_glsl_extension_directive_midshader=true` and/or `force_glsl_extensions_warn=true`.
+For Linux users running AMD GPUs, some shader packs can be fixed by launching the game with the environment
+variables `allow_glsl_extension_directive_midshader=true` and/or `force_glsl_extensions_warn=true`.
 
 Other shaderpacks aren't officially supported currently.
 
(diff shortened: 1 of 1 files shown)


coderbot16:
I'm curious how this helps for SEUS Renewed, since SEUS Renewed doesn't appear to contain GLSL #extension directives which these flags affect?

satyrnsstuff:
`allow_glsl_extension_directive_midshader=true` might not entirely necessary, although I've heard it fixes some errors with broken shaders. But, SEUS Renewed completely fails to compile without `force_glsl_extensions_warn=true` for some reason on Linux with an AMD GPU. As to why this is the case, I'm not entirely sure. It also is require to run SEUS Renewed in the same environment, with Optifine.

satyrnsstuff:
Considering the fact that this is both for Iris, and OptiFine, I will be closing this PR.

AlsoNiko:
I have had another user report this actually fixing SEUS renewed for them (but breaking complementary)
`force_glsl_extensions_warn=true`
logs for before and after are included

SEUS applied first, has tons of visual bugs. Comp is second, works fine 
[latest (32).log](https://github.com/IrisShaders/Iris/files/6836806/latest.32.log) (Seus first, than comp)

Comp first, fall back to vanilla. SEUS after, works perfectly 
[latest (33).log](https://github.com/IrisShaders/Iris/files/6836809/latest.33.log)

coderbot16:
Thanks for this PR, but it is no longer needed. Iris now moves `#extension` directives around itself to fix this issue, now these shaders should work out of the box without additional configuration with the latest commits to Iris.

PR #426, opened 2021-06-24
Update fr_fr.jsion

toinouH:
I added the missing content and corrected some of my previous translations. They weren't wrong but not perfect.

### DIFF
--- a/src/main/resources/assets/iris/lang/fr_fr.json
+++ b/src/main/resources/assets/iris/lang/fr_fr.json
@@ -24,5 +24,5 @@
   "options.iris.shaders.enabled": "Shaders : Activés",
 
   "pack.iris.select.title": "Sélectionner",
-  "pack.iris.list.label": "+ Faites glisser puis déposer vos packs de shaders pour les ajouter"
+  "pack.iris.list.label": "+ Glissez-déposez vos packs de shaders ici pour les ajouter"
 }
--- a/src/main/resources/assets/iris/lang/fr_fr.json
+++ b/src/main/resources/assets/iris/lang/fr_fr.json
@@ -1,23 +1,28 @@
 {
   "iris.shaders.reloaded": "Shaders rechargés !",
+  "iris.shaders.toggled": "Shaders basculés sur %s !",
   "iris.keybind.reload": "Recharger les shaders",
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/iris/lang/fr_fr.json
```diff
@@ -1,23 +1,28 @@
...
   "options.iris.shaderPackSelection.failedAddSingle": "\"%s\" n'est pas un shader pack valide",
   "options.iris.shaderPackSelection.copyError": "Erreur lors de l'ajout du shader pack \"%s\"",
-  "options.iris.shaderPackSelection.copyErrorAlreadyExists": "\"%s\" est déjà dans votre dossier de shader packs !",
+  "options.iris.shaderPackSelection.copyErrorAlreadyExists": "\"%s\" est déjà dans votre dossier de packs de shaders !",
   "options.iris.shaders.disabled": "Shaders : Désactivés",
   "options.iris.shaders.enabled": "Shaders : Activés",
 
   "pack.iris.select.title": "Sélectionner",
-  "pack.iris.list.label": "+ Faites glisser vos shader packs pour les ajouter"
+  "pack.iris.list.label": "+ Faites glisser puis déposer vos packs de shaders pour les ajouter"
```
maxigator, line 27: It's a weird way of saying that this would be better
```suggestion
  "pack.iris.list.label": "+ Faites glisser vos packs de shaders ici pour les ajouter"
```

maxigator (changes requested):

src/main/resources/assets/iris/lang/fr_fr.json
toinouH: I know but I checked the English version "pack.iris.list.label": "+ Drag and Drop Shader Packs to add" and it is basicly the famous "Glisser déposer" so idk if you care about keeping that parity with the English version or not.

maxigator: In that case it should be "Glissez déposez vos packs de shaders ici" because right now it looks like you have to do 2 things

toinouH: That's what we want yeah, thank you.

coderbot16 (commented):
Thanks!

PR #429, opened 2021-06-26
Issue forms for Iris

altrisi:
This PR adds some of the brand new [Github Issue Forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/) for the Iris repo.

Issue forms are similar to issue templates in a way, but they are a lot more end-user friendly, since they have proper inputs, textboxes, helper texts, placeholders, selections and checkboxes, and everything gets converted to a nice issue when being submitted.

Not only that, they also allow making some fields required, which is something quite useful IMO to prevent people just skipping the whole template and just writing something like "the game crashes" at the bottom. It also doesn't include the helper texts when the issue is submitted, only the titles and content, making it faster to read even if the user forgot to remove those parts of the templates.

This PR converts the bug report and crash templates into issue forms, having more or less the same fields those had.

You can test those issue forms on the [issue tracker of my Iris fork](https://github.com/altrisi/Iris/issues/new/choose), feel free to spam it with tests. Github also provides some kind-of preview if in the Files Changed tab you click on the `View File` button in the top right.

Feel free to mark as ready and merge it as it currently is. The reason this is a draft is because the current log/crash report section doesn't support attaching files, just pasting, since it seems there is a limitation if the `render` attribute is set (see and upvote github/feedback#4345 and/or github/feedback#4291). It can, however, be changed to a regular textarea if desired.

### DIFF
--- a/src/main/java/net/coderbot/iris/uniforms/HardcodedCustomUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/HardcodedCustomUniforms.java
@@ -21,15 +21,14 @@ public class HardcodedCustomUniforms {
 	}
 
 	private static float getTimeAngle() {
-		float sunAngle = CelestialUniforms.getSunAngle();
-		float tAmin = frac(sunAngle - 0.033333333f);
-		float tAlin = tAmin < 0.433333333f ? (tAmin * 1.15384615385f) : (tAmin * 0.882352941176f + 0.117647058824f);
-		float hA = tAlin > 0.5f ? 1.0f : 0.0f;
-		float tAfrc = frac(tAlin * 2.0f);
-		float tAfrs = tAfrc*tAfrc*(3.0f-2.0f*tAfrc);
-		float tAmix = hA < 0.5f ? 0.3f : -0.1f;
+		return getWorldDayTime() / 24000F;
+	}
 
-		return (tAfrc * (1.0f-tAmix) + tAfrs * tAmix + hA) * 0.5f;
+	private static int getWorldDayTime() {
+		long timeOfDay = MinecraftClient.getInstance().world.getTimeOfDay();
+		long dayTime = timeOfDay % 24000L;
+
+		return (int) dayTime;
 	}
 
 	private static float getTimeBrightness() {


NoComment1105 (approved):
Personally i think this is better then the templates i added a couple weeks ago because still ive seen people disregarding the template :/, just a question though, is there not one for feature requests or did you just choose to leave it out?

altrisi:
I left feature requests since I didn't think they would really benefit of a form, I think they are already nice as templates and not an obligation of how to make them, plus there's not really much extra info needed for a feature request (in a sense of making it required or using some of these new features).

I can add a form for them if you think it's worth it.

NoComment1105:
No it's fine, just wondering

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,75 @@
...
+    attributes:
+      label: Relevant log output
+      description: Please copy and paste any relevant log output. This will be automatically formatted into code, so no need for backticks.
+      render: shell
+  - type: dropdown
+    id: edition
+    attributes:
+      label: Edition
+      description: What edition of Iris are you running?
+      options:
```
yitzy299, line 39: So I'd suggest the following here:

### Iris Version

*String* 

Do you have sodium installed?

*yes | no*

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,75 @@
...
+      required: true
+  - type: input
+    id: gpu
+    attributes:
+      label: What is your GPU?
+      description: Please write the vendor and model of your GPU
+      placeholder: ex. Nvidia GeForce RTX 2070
+    validations:
+      required: true
+  - type: textarea
```
yitzy299, line 69: Should have Java version here

.github/ISSUE_TEMPLATE/crash_report.yml
```diff
@@ -0,0 +1,96 @@
...
+      value: |
+        1. Place a Redstone Lamp in front of a Redstone Repeater
+        2. Use a Lever to activate the Redstone Repeater
+        3. Game Crashes
+    validations:
+      required: true
+  - type: textarea
+    id: logs
+    attributes:
+      label: Crash Report file or latest.log
```
yitzy299, line 66: So for crash reports both the log and crash report will be useful

yitzy299 (changes requested):

coderbot16:
This PR seems to be broken? It has zero commits in it.

yitzy299:
THe bot seemed to force push removing them...

altrisi:
Crap I forgot to disable the fork updater...

Will make another one at some point.

PR #432, opened 2021-06-26
Fix world time on complementary

### LINKED ISSUE #344: Time jumps from 23200 to a weird time state until 24000
Condition: 1.16.5 Iris + Sodium + Complementary

It's hard to describe the problem without screenshots.

Here I am watching the sunrise at time 23200:
image

Here is what I'm seeing exactly two seconds after the first screenshot:
image

IMS212:
This seemingly fixes #344 without breaking BSL.


coderbot16:
Thanks for the PR, but could you target trunk instead of 1.17? I prefer to merge things forward from trunk if possible.

IMS212:
Sure

IMS212:
Will need to close this to do so, since I have 1.17 code in the commits

PR #439, opened 2021-06-27
Issue forms for Iris

altrisi:
Replaces #429. _Github still saves the commits after force pushes, so it's basically the same content._

This PR adds some of the brand new [Github Issue Forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/) for the Iris repo.

Issue forms are similar to issue templates in a way, but they are a lot more end-user friendly, since they have proper inputs, textboxes, helper texts, placeholders, selections and checkboxes, and everything gets converted to a nice issue when being submitted.

Not only that, they also allow making some fields required, which is something quite useful IMO to prevent people just skipping the whole template and just writing something like "the game crashes" at the bottom. It also doesn't include the helper texts when the issue is submitted, only the titles and content, making it faster to read even if the user forgot to remove those parts of the templates.

This PR converts the bug report and crash templates into issue forms, having the same or more fields than those had.

You can't fully test those in my repo since then I'd need to place them on my main branch and the bot would corrupt them again, but you can preview them by clicking the `View file` button of each of them in the files changed tab ([bug report](https://github.com/altrisi/Iris/blob/altrisi-issue-forms/.github/ISSUE_TEMPLATE/bug_report.yml) | [crash report](https://github.com/altrisi/Iris/blob/altrisi-issue-forms/.github/ISSUE_TEMPLATE/crash_report.yml)).

Feel free to mark as ready and merge it as it currently is. The reason this is a draft is because the current log/crash report section doesn't support attaching files, just pasting, since it seems there is a limitation if the `render` attribute is set (see and upvote github/feedback#4345 and/or github/feedback#4291). It can, however, be changed to a regular textarea if desired.

<details>
<summary>Now, what happened to the other PR?</summary>

Well, yes. It seems like I forgot that the bot I use to update my fork from upstream makes a forcepush replacing whatever there was in the main branch. That's usually not a problem, since my PRs always come from new branches, but since you can't test issue templates or forms if they are not present in the main branch, I had to do so and completely forgot to change it or disable the bot.

However, this shouldn't have been (that much of) a problem. It seems like Github instead of checking whether all commits were merged into the target branch checks whether there are no commits left in the PR's branch that are not present in the target. Therefore, since there were no commits at all, there also were no commits left not present in the target and assumed it was merged. I think that's a bug on github, since doing that is as simple as force-pushing the branch to upstream and it will be considered a "merged" PR. Whatever. In this one it won't happen again since it's not from main branch.

~Coming soon on my channel 100% reliable dupe bug how to dupe merged PRs on github~

</details>


NoComment1105 (approved):

coderbot16:
Thanks!

PR #445, opened 2021-06-28
Updated pt_BR translation for trunk

chriexpe:


coderbot16 (changes requested):
Though the majority of the changes here are nice, I'm concerned about the semantic differences from replacing refernces to "Shader Packs" with just "Shaders." Personally I'd like to keep the usage of Shader Packs consistent with the English translation, but if there's a good reason to drop the "Packs" part then I'd be willing to hear it.

IMS212:
Thanks, however this has been superseded by #888.

PR #446, opened 2021-06-28
Updated pt_BR translation for 1.17

chriexpe:


coderbot16:
Thanks for the PR, but we only need translation PRs to be targeted at the trunk branch, and then we can merge them forward. Closing this in favor of #445.

PR #448, opened 2021-06-29
Add missing translations to `de_de.json`

phibr0:
There were a few missing strings in the german translation file.


coderbot16 (commented):
Thanks!

Issue #451, opened 2021-06-29
Iris should remove whitespace before the #version directive on macOS


coderbot16:
If there's a space on the same line as the #version directive before the actual #version directive, the macOS driver fails to compile it. Just a weird Apple driver bug.

AlsoNiko:
Open PR for a fix, should be reviewed soon

coderbot16:
This was fixed by https://github.com/IrisShaders/Iris/pull/452

PR #452, opened 2021-06-29
Trim Whitespace on #version Directives

juliand665:
This addresses #451, fixing complementary in non-overworld dimensions on macOS.

As an alternative, I considered simply outputting only trimmed lines, but that made it harder to read the output of the patched shaders' source, so I chose this solution since it has more limited effects.

It also fixes Iris not recognizing `#version` directives with leading whitespace as the spot to insert its `MC_RENDER_QUALITY`/`MC_SHADOW_QUALITY` defines, though I'm not sure what effect this had, if any.


coderbot16 (approved):
Thanks! Nice work finding & fixing this issue.

PR #463, opened 2021-06-30
updated for the 1.17 release with sodium fork

laralove143:


README.md
```diff
@@ -42,7 +42,7 @@ Iris doesn't support Forge. This is for a few reasons:
 
 ## What's the current state of development?
 
-Iris has a public stable release for 1.16 that works with a custom version of Sodium, and a public beta release for 1.17 that currently does not work with Sodium (currently, there is no official version of Sodium for 1.17). Iris is still in heavy development and gets new improvements every week, and is progressing very rapidly!
+Iris has a public stable releases for 1.16 and 1.17 that work with a custom version of Sodium. Iris is still in heavy development and gets new improvements every week, and is progressing very rapidly!
```
NoComment1105, line 45: It says "that work with"
Could you update it to "that works with"

NoComment1105 (changes requested):

README.md
DestinyTheGreat: Iris has public stable releases for 1.16 and 1.17 that work with custom versions of Sodium. Iris is still in heavy development and gets new improvements every week, and is progressing very rapidly!

This would be grammatically correct

DestinyTheGreat (changes requested):
As well, Iris for 1.17 is not entirely stable, so for now I would not change the README

laralove143:
why not mention both with a public stable release for 1.16 and a public unstable release for 1.17...

RDKRACZ:
bump

coderbot16:
Thanks for the PR, I have updated the Current State section accordingly myself: https://github.com/IrisShaders/Iris/commit/065a0d6d99e2fcb460c5d59297b4fd91c1a5dbee

PR #467, opened 2021-07-01
Estonian translation

Madis0:
One thing I didn't find is %s variants for `iris.shaders.toggled`, where are they?


coderbot16 (approved):
Thanks!

coderbot16:
> One thing I didn't find is %s variants for iris.shaders.toggled, where are they?

Currently "off" is hardcoded to English. Otherwise, the %s will be the name of the shader pack.

coderbot16:
https://github.com/IrisShaders/Iris/issues/813

PR #480, opened 2021-07-03
Fix 1.17.1 compatibility

### LINKED ISSUE #478: Can you add 1.17.1 snaps support
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

### LINKED ISSUE #479: I know it is for 1.17 and not 1.17.1 pre or 1.17.1 yet but..
I asked for a new version for the mod I was told that non releases / betas are not given to people (probably non patrons )
But I tried "updating" it myself and what I mean allowing the other version to be able to be used.
I was told that it should just work but then when I ran it I realized that it won't and I think It has to do with the patching of the render glitch from 1.17, so I guess you don't give it out but You can fix it by changing 
something to do with alwaysRenderSky or / from class_315 and it showe dfrom this mixin, mixins.iris.json:MixinWorldRenderer
It failed to inject multiple times so Yeah thx bye.

IMS212:
Closes #478 
Closes #479

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
+++ b/src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
@@ -91,12 +91,6 @@ public class MixinWorldRenderer {
 		pipeline.setPhase(WorldRenderingPhase.SKY);
 	}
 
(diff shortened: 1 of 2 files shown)


JackTGG:
oh nice

PR #481, opened 2021-07-03
Support 1.17 lines rendering for old shaderpacks

### LINKED ISSUE #403: Lines shader not working in 1.17
Repro steps:

Launch Iris 1.17 stand alone installation
Enter any world
Enable shaders: BSL or Complementary (latest dev versions, all default settings)
Expected: Default black selection outline should appear where crosshair is pointing
Observed: No selection outline at all
Note: This could be an issue with other shaders too, only tested using BSL and Comp at this point]

CodingRays:
Added support for old shaders to support the new 1.17 lines rendering. This is done by injecting code into the lines vertex shader that transforms the vertices before the actual shader. To easily support newer shaders that do have proper lines handling this is only done if the legacyLines flag is set in the ProgramSet. By doing this no changes in any minecraft render code is needed which hopefully makes it easier to maintain support for this going forward.

There are still some issues.
1. ~~Line width is calculated in world space making lines change size depending on their distance to the camera. I have some ideas of how to do this in screen space as well by performing the calculations after the shader has run but there are many more issues with this that make it much harder to implement.~~
~~Line width is calculated in ndc space. Minceraft doe this in screen space however as far as im aware we have no uniforms providing screen size information.~~
2. ~~There is currently no support for the minecraft provided line width uniform.~~ (Thanks IMS)
3. The Transformations class could use a refractor.
4. Better way to tell the `TriforcePatcher::patch` function to inject the code than passing a boolean through 3 other functions?

Closes #403


IMS212:
Is this ready for review?

CodingRays:
Yes

src/main/java/net/coderbot/iris/shaderpack/transform/Transformations.java
```diff
@@ -11,6 +11,7 @@
 	enum InjectionPoint {
 		DEFINES,
 		BEFORE_CODE,
+        MAIN_TAIL,
```
IMS212, line 14: Wrong formatting

src/main/java/net/coderbot/iris/pipeline/newshader/TriforcePatcher.java
```diff
@@ -193,6 +194,37 @@ public static String patch(String source, ShaderType type, AlphaTest alpha, bool
...
+		// Inject the legacy lines transformation after everything else
+		if(type == ShaderType.VERTEX && injectLegacyLines) {
+			transformations.injectLine(Transformations.InjectionPoint.BEFORE_CODE, "uniform float iris_LineWidth;");
+			// TODO: Need a better way to do this
+			transformations.replaceExact("uniform vec3 cameraPosition;", ""); // Required if shader has cameraPosition inside a guard
+			transformations.injectLine(Transformations.InjectionPoint.BEFORE_CODE, "uniform vec3 cameraPosition;");
+
+			transformations.replaceExact("uniform float viewWidth;", "");
+			transformations.replaceExact("uniform float viewHeight;", "");
+			transformations.replaceExact("uniform float viewWidth, viewHeight;", ""); // This is sooo scuffed
```
IMS212, line 206: It would be better to avoid comments like these, and replace them with more useful info.

src/main/java/net/coderbot/iris/pipeline/newshader/TriforcePatcher.java
```diff
@@ -193,6 +194,37 @@ public static String patch(String source, ShaderType type, AlphaTest alpha, bool
...
+					"vec3 _iris_internal_ndcStart = _iris_internal_ndcStart4.xyz / _iris_internal_ndcStart4.w;\n" +
+					"vec3 _iris_internal_ndcEnd = _iris_internal_ndcEnd4.xyz / _iris_internal_ndcEnd4.w;\n" +
+					"vec2 _iris_internal_screenSize = vec2(viewWidth, viewHeight);\n" +
+					"vec2 _iris_internal_lineDir = normalize(_iris_internal_ndcEnd.xy - _iris_internal_ndcStart.xy);\n" +
+					"vec2 _iris_internal_lineOffset = (vec2(-_iris_internal_lineDir.y, _iris_internal_lineDir.x) * (iris_LineWidth / _iris_internal_screenSize));\n" +
+					"if(_iris_internal_lineOffset.x < 0) { _iris_internal_lineOffset = _iris_internal_lineOffset * -1; }\n" +
+					"if(gl_VertexID % 2 == 0) {\n" +
+					"gl_Position = (vec4(_iris_internal_ndcStart, 1.0) + vec4(_iris_internal_lineOffset, 0.0, 0.0) - vec4(0.0, 0.0, 0.0001, 0.0)) * _iris_internal_ndcStart4.w;\n" +
+					"} else {\n"+
+					"gl_Position = (vec4(_iris_internal_ndcStart, 1.0) - vec4(_iris_internal_lineOffset, 0.0, 0.0) - vec4(0.0, 0.0, 0.0001, 0.0)) * _iris_internal_ndcStart4.w;\n" +
```
IMS212, line 222: Is there a reason for this duplicate code?

IMS212 (changes requested):
Looks mostly good to me, however a few changes should be made.

src/main/java/net/coderbot/iris/shaderpack/transform/Transformations.java
CodingRays: Fixed. (And im gonna blame this on intellij which displayed it properly)

src/main/java/net/coderbot/iris/pipeline/newshader/TriforcePatcher.java
CodingRays: Its not duplicate. Every other vertex must be pushed in the opposite direction. Otherwise the line would not have any width.

IMS212: Oh, didn't see the difference, my bad.

src/main/java/net/coderbot/iris/pipeline/newshader/TriforcePatcher.java
CodingRays: Is this better now?

IMS212: Yep.

IMS212:
Looks good to me.

IMS212 (approved):

TehcJS (approved):

coderbot16:
This code is a good start, and seems to work well enough for common cases. However, it's a bit more fragile than I'd like, and has the potential to cause bugs. For example, this does not work with BSL's TAA jitter or world curvature it seems. I'm investigating to see if there's a way to not completely override the shaderpack's position calculations.

TehcJS:
> This code is a good start, and seems to work well enough for common cases. However, it's a bit more fragile than I'd like, and has the potential to cause bugs. For example, this does not work with BSL's TAA jitter or world curvature it seems. I'm investigating to see if there's a way to not completely override the shaderpack's position calculations.

bend the line vertices to fit the block vertices in `gbuffers_terrain.vsh` and `gbuffers_block.vsh`
boom fixed

CodingRays:
I originally was modifying the vertices before the shader. But that was quiet unstable and made performing the calculations in screen space very slow. My older commits have the code infrastructure necessary for it, with the now fixed shader code it should be possible to get it more stable. But still very slow.

Specifically commit 5981fe2 moved the update to the end of the shader.

IMS212:
This still seems quite broken. (Complementary 4.2 development version 2+ fails to compile at all, with a supposedly invalid main function.)

coderbot16:
Thanks for the PR. I did up finding a different approach that seems to work better with most packs, so I'll be closing this PR, but this code was definitely useful as an initial reference for that approach.

PR #505, opened 2021-07-10
Update for Minecraft 1.17.1

WhyDoWeWonder:
Updates Iris to build for and run on 1.17.1, tested with iris-sodium, no issues whatsoever.


IMS212:
You seem to be including all of Fabric API as a dependency for seemingly no reason. Also, any reason you're using 0.36.1? As far as i know the only thing that accomplishes is breaking backward compatibility.

WhyDoWeWonder:
Yeah that's my bad, I already had 0.36.1 downloaded as a jar and didn't want to download another API version

IMS212:
Already done in recent commits.

PR #538, opened 2021-07-14
Translate mojang's overlay colors to replace entityColor

IMS212:
This adds code to the vertex shader that gets the appropriate vec4 from UV1 and the sampler2d "overlay" (previously Sampler1), and passes it to the fragment shader. (This also replaces all instances of "entityColor.a" with "1 - entityColor.a", as for some unknown reason not doing this causes the entity to become a solid color.)


IMS212:
Due to how this works, this is completely incompatible with 1.16.

PR #549, opened 2021-07-15
Capture current atlas for use with the atlasSize uniform

IMS212:
I haven't been able to confirm this is completely correct, but it seems to work fine testing in Complementary.
In 1.17 we will likely need to mixin into setShaderTexture in RenderSystem instead.


coderbot16:
This doesn't appear to have any code to handle the absence of a currently bound atlas, which would make atlasSize be (0, 0). Also, it doesn't appear to have any handling to check whether or not the atlas is being bound to texture unit 0.

I'll take a look at this myself.

PR #568, opened 2021-07-17
Add enableLogging option and enable by default in a development environment

IMS212:


coderbot16:
I imagine that most of the problematic logging was either fixed or removed leading up to 1.1.2, not sure if this is still needed. I'm not really a fan of disabling logging outright - either a message should be kept because it's helpful, or removed because it's wrong or useless.

IMS212:
This is useless now that an alternative is upstream.

PR #572, opened 2021-07-18
Update README.md Added A Brillient Shader !

samdhi2323:


HiItsDevin:
You should probably link their official CurseForge/Website page, and not a reupload. This is their official site: [https://dedelner.net/kuda-shaders/](https://dedelner.net/kuda-shaders/)

yitzy299:
You’ll need to link the correct site, not one known for illegal reuploads and viruses, and “I think this works” isn’t what we want in the readme

PR #590, opened 2021-07-19
Fix dirty issue by ignoring macOS files (like .DS_Store)

douira:
macOS generates `.DS_Store` and other files automatically in many folders when Finder is used. Ignoring them in gitignore resolves the issue where builds on macOS are considered dirty. (Running `git status` will reveal the presence of an untracked `.DS_Store`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,9 @@ bin/
 
 run/
 /logs/latest.log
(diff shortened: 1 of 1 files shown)


IMS212:
Thanks!

PR #594, opened 2021-07-20
Fix memory leak

IMS212:


src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
```diff
@@ -45,18 +50,28 @@
 	private static final String RENDER_WEATHER = "Lnet/minecraft/client/render/WorldRenderer;renderWeather(Lnet/minecraft/client/render/LightmapTextureManager;FDDD)V";
 	private static final String RENDER_WORLD_BORDER = "Lnet/minecraft/client/render/WorldRenderer;renderWorldBorder(Lnet/minecraft/client/render/Camera;)V";
 
+	@Unique
+	HorizonRenderer horizonRenderer = new HorizonRenderer();
```
coderbot16, line 54: This can probably be private?

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/mixin/MixinWorldRenderer.java
IMS212: Sure, didn't think of much reason to keep it private but again, not too familiar with accessing.

IMS212: Fixed.

coderbot16: Yeah, it's best to keep fields and similar things private unless they need to be public, that's part of encapsulation.

coderbot16 (approved):
Thanks for finding this fix! Hopefully this resolves the remaining performance collapse & memory leak issues on 1.17.

PR #597, opened 2021-07-21
Remove internal shaders

IMS212:
You cannot use them as of now, so having the shaders themselves be included doesn't make much sense.


coderbot16 (approved):
🦀

PR #601, opened 2021-07-21
Merge latest changes from trunk

IMS212:
Currently, the new fog uniforms are commented out due to fog density and mode not existing as a concept in 1.17.


IMS212:
I believe this is now good enough to merge, however a final review would be good. Also, fog uniforms are currently still commented out with a TODO.

PR #605, opened 2021-07-22
Fix shaders using the wrong time when in a dimension with fixed time

### LINKED ISSUE #547: Incorrect time in dimensions with fixed time
### What happened?

In a dimension with a fixed time the shader will use the time in the overworld instead of the time of the dimension you are in. This can result in night time in a perma-day world.

To reproduce: 
1. Create a new world and install the attached datapack ( [fixed_time.zip](https://github.com/IrisShaders/Iris/files/6823263/fixed_time.zip) )
2. Go to the alternative dimension by running `/execute in suppergerrie2:fixed_time run tp @s ~ 64 ~`
3. Run `/time set night` and `/time set day` both with shaders enabled and disabled
4. Observe how with shaders disabled the sun does not change but with shaders enabled the sky changes based on the overworld time.

### Screenshots

_No response_

### Relevant log output

_No response_

iris-mc1.17-1.1.0+rev.01d8631 (latest commit on 1.17 branch at time of writing)

No

### Operating System

Windows 10

### What is your GPU?

NVIDIA GeForce RTX 2060

### Java Version

Java 16

### Additional context

The issue is caused by [WorldTimeUniforms#getWorldDayTime](https://github.com/IrisShaders/Iris/blob/01d8631f1887a055a8305d9aef3eb93f69e29364/src/main/java/net/coderbot/iris/uniforms/WorldTimeUniforms.java#L28-L33) and [HardcodedCustomUniforms#getWorldDayTime](https://github.com/IrisShaders/Iris/blob/01d8631f1887a055a8305d9aef3eb93f69e29364/src/main/java/net/coderbot/iris/uniforms/HardcodedCustomUniforms.java#L27-L32).

These need to be fixed to take the fixed time into account. The only way I've found to get the fixed time is by adding an accessor for DimensionType.fixedTime and then changing getWorldDayTime to check that:

```java
((DimensionTypeAccessor) world.getDimension()).getFixedTime().orElse(timeOfDay % 24000L);
```

suppergerrie2:
Fixes #547 by taking into account the fixed time of the dimension.

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/DimensionTypeAccessor.java
+++ b/src/main/java/net/coderbot/iris/mixin/DimensionTypeAccessor.java
@@ -1,0 +1,15 @@
+package net.coderbot.iris.mixin;
+
+import net.minecraft.world.dimension.DimensionType;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+
+import java.util.OptionalLong;
+
+@Mixin(DimensionType.class)
+public interface DimensionTypeAccessor {
+
+	@Accessor
+	OptionalLong getFixedTime();
+
+}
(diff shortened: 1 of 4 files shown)


coderbot16:
Does this apply to 1.16? If so, we'd prefer for PRs to be targeted against `trunk` so that they can be merged forward.

suppergerrie2:
It applies to both 1.16 and 1.17

coderbot16 (approved):
Thanks!

PR #624, opened 2021-07-24
Physics Mod compatibility, alongside mod loading checks.

IMS212:
This changes the behavior of Iris to only check for Sodium/PhysicsMod once while the game loads and store it in a static variable, and not to check it during gameplay. This also changes the LivingEntityRenderer mixin to not activate a red tint when an entity is about to die if Physics Mod is installed, as not doing this causes unintended side effects.

### DIFF
--- a/src/main/java/net/coderbot/iris/Iris.java
+++ b/src/main/java/net/coderbot/iris/Iris.java
@@ -43,9 +43,9 @@ public class Iris implements ClientModInitializer {
 
 	private static ShaderPack currentPack;
 	private static String currentPackName;
-	public static boolean internal;
-	public static boolean sodiumInstalled;
-	public static boolean physicsModInstalled;
+	private static boolean internal;
+	private static boolean sodiumInstalled;
+	private static boolean physicsModInstalled;
 
 	private static PipelineManager pipelineManager;
 	private static IrisConfig irisConfig;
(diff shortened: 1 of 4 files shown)


IMS212:
Turns out there are still some issues. Disabling and re-enabling shaders causes tons of logspam, of OpenGL 1282 errors stating the array object is inactive.

IMS212:
These issues have been fixed on Physics Mod's side, and this should be ready for review.

IMS212:
Marking this as high priority as without this PR, Physics mod compatibility will break in the next update to Iris.

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -43,7 +43,9 @@
 
 	private static ShaderPack currentPack;
 	private static String currentPackName;
-	private static boolean internal;
+	public static boolean internal;
```
coderbot16, line 46: This has no need to be public.

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -43,7 +43,9 @@
...
+	public static boolean sodiumInstalled;
+	public static boolean physicsModInstalled;
```
coderbot16, line 48: These should be private, but have getter methods.

coderbot16 (changes requested):
Looks good aside from a few minor encapsulation issues.

IMS212:
All of the issues you brought up should be fixed in the newest commit.

coderbot16 (approved):
Thanks!

PR #626, opened 2021-07-25
Add glint support

IMS212:
This also splits the composite rendering and pipeline end into 2 separate injects, and has the pipeline end truly be at the end of rendering.


IMS212:
We've found a way to accomplish this without splitting the world renderer mixin, by injecting with a shift into RETURN.

IMS212:
This should be ready for merge.

coderbot16 (approved):
Thanks!

PR #634, opened 2021-07-26
Allow Sodium to store block entity information about the shadow pass

IMS212:
This will require a small change to Sodium as well.

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/shadows/MixinPreventRebuildNearInShadowPass.java
+++ b/src/main/java/net/coderbot/iris/mixin/shadows/MixinPreventRebuildNearInShadowPass.java
@@ -1,11 +1,15 @@
 package net.coderbot.iris.mixin.shadows;
 
+import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import net.coderbot.iris.pipeline.ShadowRenderer;
+import net.minecraft.block.entity.BlockEntity;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.render.Camera;
 import net.minecraft.client.render.Frustum;
 import net.minecraft.client.render.WorldRenderer;
+import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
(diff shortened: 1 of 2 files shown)


src/main/java/net/coderbot/iris/mixin/shadows/MixinPreventRebuildNearInShadowPass.java
```diff
@@ -28,4 +36,14 @@
...
 	}
+
+	@Inject(method = "setupTerrain", at = @At("TAIL"))
+	private void iris$copyBlockEntities(Camera camera, Frustum frustum, boolean hasForcedFrustum, int frame, boolean spectator, CallbackInfo ci) {
+		ShadowRenderer.visibleBlockEntities.clear();
+		for (WorldRenderer.ChunkInfo chunk : this.visibleChunks) {
+			for (BlockEntity entity : ((ChunkInfoAccessor) chunk).getChunk().getData().getBlockEntities()) {
+				ShadowRenderer.visibleBlockEntities.add(entity);
+			}
+		}
```
coderbot16, line 47: This should check whether the shadow pass is active before computing visibility info.

coderbot16 (changes requested):

coderbot16:
I think the ShadowRenderer.visibleBlockEntities list should be set to null at the end of shadow rendering to ensure we aren't holding on to (and leaking) a bunch of block entities if the user switches away from a shadowmapped shader pack. The list can be recreated each frame perhaps instead.

IMS212:
For some unknown reason, doing a check for if the shadow renderer is active seems to cause no block entity shadows to render at all.

coderbot16:
That's because the injection is at TAIL, but in the shadow pass setupTerrain is cancelled before rebuildNear. Might as well move the code to the pre-rebuildNear injection I guess.

IMS212:
Makes sense, i'll attempt a workaround for this.

src/main/java/net/coderbot/iris/mixin/shadows/MixinPreventRebuildNearInShadowPass.java
IMS212: Fixed

IMS212:
This should be fixed.

coderbot16 (approved):
Thanks!

PR #660, opened 2021-07-29
Expose geometry shaders to SodiumTerrainPipeline

IMS212:


TehcJS (approved):
Looks good to me.

IMS212:
Will merge soon if there are no complaints, as this shouldn't cause issues being only the Iris interface and not Sodium side.

PR #667, opened 2021-07-30
Deduplicate Sodium version references

msfjarvis:
The required Sodium version was changed to 0.3.0-IRIS2 but the error string incorrectly still mentioned 0.3.0-IRIS1, causing quite a bit of lost time figuring out why my client wasn't working.


IMS212:
Thanks!

msfjarvis:
> Thanks!

Thanks for fixing the typo, not sure how I missed it 😅

PR #669, opened 2021-07-30
Readd fog density on 1.17

IMS212:


src/main/java/net/coderbot/iris/mixin/statelisteners/MixinGlStateManager.java
```diff
@@ -36,8 +36,8 @@
 	}
 
 	static {
```
TehcJS, line 38: just delete the mixin

src/main/java/net/coderbot/iris/uniforms/CameraUniforms.java
```diff
@@ -27,7 +27,7 @@ public static void addCameraUniforms(UniformHolder uniforms, FrameUpdateNotifier
 			.uniform3d(PER_FRAME, "previousCameraPosition", new PreviousCameraPosition(notifier));
 	}
 
-	private static int getRenderDistanceInBlocks() {
+	public static int getRenderDistanceInBlocks() {
 		return client.options.viewDistance * 16;
```
TehcJS, line 31:
```diff
- return client.options.viewDistance * 16;
+ return client.gameRenderer.getViewDistance();
```
Use `GameRenderer#getViewDistance`

TehcJS (changes requested):
:bike::house:

src/main/java/net/coderbot/iris/uniforms/CameraUniforms.java
IMS212: this should be something for another pr, as this is the current behavior of CameraUniforms.

IMS212:
This doesn't seem to work properly, closing for now.

PR #678, opened 2021-07-31
Reduce icon file size.

RDKRACZ:
The old logo was 290x290 and its size was 18,9 KB but the current one is 1280x1280 - 156,2 KB. This pr resizes the icon to 256x256 (Fabric loader doc says that a power of two is recommended) and compresses it with no visible visual differences. The new icon file size is 9,4 KB.

I also made a version that is two times smaller than the original one so 640x640 and weights 23,1 KB.

If you do not want this one-line pr then just download the modified icon from my fork or I can just send it here.


coderbot16:
Thanks for the PR! I've used an updated base file for the logo that removes from artifacts as well as the service https://compress-or-die.com, in order to make the logo even smaller - only 7 kB at 256x256.

PR #680, opened 2021-08-01
add working fogStart and fogEnd

### LINKED ISSUE #681: there is no fogStart and fogEnd
### What happened?

A bug happened!

### Screenshots

_No response_

### Relevant log output

_No response_

iris 1.1.1 and 1.1.2-pre 1.17.1

No

### Operating System

ubuntu 20.04

### What is your GPU?

absolutely not

### Java Version

java -1

### Additional context

_No response_

TehcJS:
fixes #681


coderbot16:
Why not use `gl_Fog.start` and `gl_Fog.end` ?

TehcJS:
glsl 1.5 removed those

coderbot16:
It's still valid to use them with Iris, since it patches from the compatibility profile to the core profile. But I understand that some shader authors really don't want to work with compatibility profile features.

I'd rather see what OptiFine is doing with regards to exposing vanilla uniforms like this to shaders instead of implementing our own incompatible approach.

TehcJS:
it is what optifine is doing though? it's in the optifine docs

coderbot16:
There are no mentions of fogStart and fogEnd in the [optifine docs](https://github.com/sp614x/optifine/blob/eeb28cf187ff8d9beb3fb5442a7631d0029fd34e/OptiFineDoc/doc/shaders.txt).

TehcJS:
is there any alternative to this without using the compatibility profile?

IMS212:
According to some info in shaderlabs, fogStart and fogEnd are valid.

TehcJS:
Also, I tested it in optifine; it's valid to use `fogStart` and `fogEnd`.

TehcJS:
image

coderbot16:
`FogStart` and `FogEnd` aren't the same as `fogStart` and `fogEnd`, uniform names are case sensitive.

TehcJS:
I can confirm `fogStart` and `fogEnd` work in the latest Optifine version 1.17.1

ColdIce1605:
Oh yeah I do remember that we should probably allow vanilla uniforms as that kinda how OF patches shaders.

ColdIce1605:
@TehcJS (not sure if you're watching)
You could change this PR to be more general and allow all built-in vanilla uniforms too of course along with the OF names(old uniforms)

TehcJS:
ok i will do that tmr or something

PR #695, opened 2021-08-03
Update to Loom 0.9, yarn build 32, and link in gradle.properties

IMS212:
This has no changes on it's own, but is good for futureproofing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-	id 'fabric-loom' version '0.8-SNAPSHOT'
+	id 'fabric-loom' version '0.9-SNAPSHOT'
 	id 'maven-publish'
 	id 'org.ajoberstar.grgit' version '4.1.0'
 }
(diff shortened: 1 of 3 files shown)


IMS212:
Will merge soon if there are no complaints, as this is a small change.

coderbot16:
Mappings updates can be dangerous in the context of Yarn's name swaps. Have you audited the mappings changes for any issues like that?

IMS212:
I haven’t audited it, however I have tested it and it runs fine in development.

IMS212:
I've just tested it in production, and it runs fine.

coderbot16 (changes requested):
Gradle & loom updates are fine, but I think we shouldn't update Yarn. Instead, we should focus on moving to mojmap. Also, is there any reason we can't use Loom 0.9 / modern Gradle on trunk?

IMS212:
Loom 0.9 forces the use of Java 16. The mojang mapping PR is almost complete, just needs a few finishing touches.

DragonEggBedrockBreaking:
As IMS said, Loom 0.9 forces use of Java 16, but modern gradle should be fine - afaik Loom 0.6 supports Gradle 7.x

BeansBeefBroccoli:
Updating yarn should no longer be needed now as coderbot and IMS have converted to Iris mojmap.

PR #696, opened 2021-08-03
Fix overlay texture unit, fixes hit flash without shaders

IMS212:
I couldn't find any reason this was set to 6, am I missing something?


coderbot16:
This was previously set to 6 to avoid a potential clash with the old sampler system, but it's no longer an issue. I'm curious why setting this to 2 instead of 6 fixes issues, however, since vanilla sets the overlay texture unit to 1, so it doesn't even match vanilla. Odd.

IMS212:
Not sure either, however it seems to work without any noticeable issues with shaders on or off.

IMS212:
This should be ready for merge.

IMS212:
Marking this as high priority, as this breaks vanilla behavior.

IMS212:
Merging, as I have been testing this for a few hours with no noticable issues.

PR #742, opened 2021-08-10
Add new Sodium incompatibility screen

IMS212:
This replaces the null pointer exception error, and looks somewhat like this:


src/main/java/net/coderbot/iris/mixin/MixinTitleScreen.java
```diff
@@ -0,0 +1,25 @@
...
+import net.minecraft.util.Formatting;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(TitleScreen.class)
+public class MixinTitleScreen {
+	@Inject(method = "init", at = @At("RETURN"))
+	public void showError(CallbackInfo ci) {
```
FoundationGames, line 18: Prefix with `iris$` and maybe make the inject method name more descriptive, i.e. "showSodiumIncompatScreen"

FoundationGames (changes requested):

src/main/java/net/coderbot/iris/mixin/MixinTitleScreen.java
IMS212: sounds good, will do this

IMS212: Done.

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -63,7 +64,7 @@ public void onInitializeClient() {
 					// A lot of people are reporting visual bugs with Iris + Sodium. This makes it so that if we don't have
 					// the right fork of Sodium, it will just crash.
 					if (!versionString.startsWith("0.2.0+IRIS3")) {
-						throw new IllegalStateException("You do not have a compatible version of Sodium installed! You have " + versionString + " but 0.2.0+IRIS3 is expected");
+						sodiumInvalid = true;
```
coderbot16, line 67: We probably still want to have some sanity check that crashes the game if the wrong version of Sodium is installed and the user tries to load a world.

coderbot16 (commented):
Looks good to me, one minor comment.

Madis0:
Where does "back" lead to anyway? Shouldn't it close the client?

src/main/java/net/coderbot/iris/Iris.java
IMS212: The user is never able to load the world with this implementation, clicking back closes the game. (I can't change what back says without creating a new Screen instance)

IMS212:
Correct, it schedules a safe close via scheduleStop()

src/main/java/net/coderbot/iris/mixin/MixinTitleScreen.java
```diff
@@ -0,0 +1,25 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(TitleScreen.class)
+public class MixinTitleScreen {
+	@Inject(method = "init", at = @At("RETURN"))
+	public void iris$showSodiumIncompatScreen(CallbackInfo ci) {
+		if(Iris.sodiumInvalid) {
+			MinecraftClient.getInstance().openScreen(new NoticeScreen(() -> {
+				MinecraftClient.getInstance().scheduleStop();
+			}, new LiteralText("Iris failed to load!").formatted(Formatting.RED), new LiteralText("You have Sodium in your mods folder, however Iris includes it. Please close the game, and remove Sodium from your mods folder.")));
```
Madis0, line 22: If possible, these could be TranslatableText.

src/main/java/net/coderbot/iris/Iris.java
Madis0: > I can't change what back says without creating a new Screen instance

Is there a way to display it without the button though? That would reduce the confusion

IMS212: Not without a new screen, I could create one if there was demand but it would just need another class.

src/main/java/net/coderbot/iris/mixin/MixinTitleScreen.java
IMS212: Will do, thanks for bringing this up.

IMS212:
I've encapsulated sodiumInvalid and made the text translatable.

src/main/java/net/coderbot/iris/Iris.java
coderbot16: > The user is never able to load the world with this implementation

Even though I too believe this to be true, it wouldn't hurt to still add a check & crash in world rendering to be 100% absolutely sure that nothing unexpected happens causing the game to proceed past the warning screen.

IMS212:
I'm going to be taking Sodium's new incompatibility screen and using that instead of vanilla's screen which I can't change, so please don't merge this yet.

src/main/java/net/coderbot/iris/Iris.java
IMS212: Will do this later.

IMS212:
That sounds good as well, I'll quickly fix the merge conflict.

IMS212:
Done, feel free to merge.

coderbot16 (approved):
Thanks! I'll add the sanity check crash myself. This should help prevent one of the more common support requests in the Iris discord.

PR #746, opened 2021-08-11
Run the IntelliJ code formatter on all code

IMS212:
This doesn't optimize imports or rearrange code, to avoid star imports. This only adds tabs and similar, to make the code look nicer, however this comes with a huge git blame sadly.


coderbot16:
Looks like this PR got obliterated by the mojmap transition. Since this was an automated process it could probably just be redone. That being said, I think the double-indentation on line continuations that IntelliJ is doing is odd, I wonder if there's a way to disable that.

PR #750, opened 2021-08-11
Add fallback path in case the atlassize didn't get stored

Kroppeb:
I tested this be commenting out the original code, and it seems to not crash with the fix. I couldn't really see a difference with noisy textures on or off so idk if it actually works as intended.

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/MixinSpriteAtlasTexture.java
+++ b/src/main/java/net/coderbot/iris/mixin/MixinSpriteAtlasTexture.java
@@ -1,11 +1,14 @@
 package net.coderbot.iris.mixin;
 
+import com.mojang.blaze3d.systems.RenderSystem;
 import net.coderbot.iris.texunits.SpriteAtlasTextureInterface;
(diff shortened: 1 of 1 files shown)


coderbot16 (changes requested):
The issue I see with this code is that it potentially trashes the current texture binding. There should be a way to restore the previous binding once the code is done.

Kroppeb:
> The issue I see with this code is that it potentially trashes the current texture binding. There should be a way to restore the previous binding once the code is done.

Oh, I thought `RenderSystem.bindTexture(glId);` would be all that is needed.

coderbot16:
Changing the texture binding with bindTexture and not restoring it is dangerous when this code could be called at pretty much any time - it would work fine most of the time but potentially cause weird intermittent issues some of the time. Restoring the binding at the end would avoid those potential issues.

coderbot16 (commented):
I'll merge this and fix the texture binding issue myself. Thanks for the dashloader compatibility fix!

PR #771, opened 2021-08-19
Dragon death progress + new code for Iris exclusive defines/uniforms

IMS212:
This implements IrisUniforms and getIrisDefines(), for future Iris-exclusive macros/uniforms. This also adds the first of what may be in the future many Iris-exclusive uniforms, and an appropriate definition for shaders to check for the existence of it.


src/main/java/net/coderbot/iris/mixin/MixinEnderDragonEntityRenderer.java
```diff
@@ -0,0 +1,23 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(EnderDragonEntityRenderer.class)
+public class MixinEnderDragonEntityRenderer {
+	@Inject(method = "render", at = @At("HEAD"))
+	private void setTicksSinceDeath(EnderDragonEntity enderDragonEntity, float f, float g, MatrixStack matrixStack, VertexConsumerProvider vertexConsumerProvider, int i, CallbackInfo ci) {
+		if (enderDragonEntity.ticksSinceDeath != 0) {
+			CapturedRenderingState.INSTANCE.setTicksSinceDragonDeath(enderDragonEntity.ticksSinceDeath);
```
coderbot16, line 18: This is a very undesirable way of doing this since the uniform won't have a valid value until the middle of the frame. Also, it doesn't handle multiple ender dragons either, it could be sane to define this as the maximum death progress of any ender dragon alive.

src/main/java/net/coderbot/iris/gl/shader/StandardMacros.java
```diff
@@ -150,6 +151,13 @@ public static String group(Matcher matcher, String name) {
 		return Arrays.stream(extensions).map(s -> "MC_" + s).collect(Collectors.toList());
 	}
 
+	public static List<String> getIrisDefines() {
+		List<String> defines = new ArrayList<>();
+		defines.add("dragonDeathProgressSupported");
```
coderbot16, line 156: We should choose a name that is more in line with the other standard macros.
```suggestion
		defines.add("MC_UNIFORM_DRAGON_DEATH_PROGRESS");
```

src/main/java/net/coderbot/iris/uniforms/IrisUniforms.java
```diff
@@ -0,0 +1,13 @@
+package net.coderbot.iris.uniforms;
+
+import net.coderbot.iris.gl.uniform.UniformHolder;
+
+import static net.coderbot.iris.gl.uniform.UniformUpdateFrequency.PER_TICK;
+
+public class IrisUniforms {
```
coderbot16, line 7: Maybe IrisExclusiveUniforms or similar?

coderbot16 (changes requested):

coderbot16:
Also, we define the progress in ticks, but wouldn't it be more desirable to use `0.0` for "death has just started" and `1.0` for "death complete / no death in progress"?

src/main/java/net/coderbot/iris/mixin/MixinEnderDragonEntityRenderer.java
IMS212: This is the method vanilla seems to use when rendering, should we just rename the uniform to be in line with this?

IMS212:
Vanilla itself defines it in ticks, so I don't think we should change it.

IMS212:
After thinking a bit, percentage would make more sense in this case.

IMS212:
Change has been made.

src/main/java/net/coderbot/iris/mixin/MixinEnderDragonEntityRenderer.java
IMS212: About the multiple dragon issue, this should never be an issue realistically.

coderbot16:
Defining it in terms of 0.0 to 1.0 makes "dragonDeathProgress" make more sense when there isn't a dragon. If it were ticks, "dragonDeathTime" would be a more sensible name.

In any case, based on the code, this will be the meaning of each progress value, at least in the current version of the game:

- `x = 0.0`: at least one dragon is alive and not dying
- `0.0 < x < 0.75`: dragon is dying
- `0.75 < x < 0.9`: dragon is dying and emitting experience orbs
- `0.9 < x < 1.0`: dragon is dying, emitting experience orbs, and exploding
- `x = 1.0`: dragon is dead, or no dragon exists

The actual value of dragonDeathProgress should be the minimum dragonDeathProgress corresponding to any single dragon. In other words, if there's a dying dragon and an alive and healthy dragon, dragonDeathProgress should probably remain as `0.0`. I don't imagine that many people will be encountering multiple dragons though.

From looking at the code, it looks like a dragon takes 200 ticks (10 seconds) to die.

IMS212:
There's no simple way to look for the existence of multiple dragons sadly, to my knowledge.

coderbot16:
Well, there's still the issue that dragon death progress doesn't have its proper value until the middle of the frame. So in the shadow pass, you'll actually be seeing the death progress of the previous frame.

IMS212:
Sadly I can’t think of any reasonable solution to this, nor would this be a major issue as it’s only updated per tick.

coderbot16:
It should be possible to figure out the list of all dragons in the world somehow. The hard part is doing that efficiently.

IMS212:
I haven’t found any clear way to do so yet.

IMS212:
After a bit of review, I still can’t find a way to look for the existence of multiple dragons, would using a min value of the current value/old tick value and then resetting the value when the death animation completes suffice?

IMS212:
Outside of the seemingly-impossible dragon check, this should be ready to merge.

coderbot16:
This PR still has a bug where dragonDeathProgress is only updated midway through the frame, after other shader programs have already ran.

IMS212:
There's no clean way to avoid this - you can't check a global value for dragon death progress.

coderbot16:
It is certainly possible to keep track of the alive dragons that are currently within the player's render distance, but it could be a bit hard to do that without introducing a small performance hit if there are a ton of entities in the world.

Also, this won't work properly if shadows are disabled and the player is not looking at the dragon for an extended period of time.

IMS212:
Ok, I think it's time to come back to this now that we have a clean state. What do you propose we do for this? This should probably be the first stage of iris-exclusive features to begin with.

PR #776, opened 2021-08-20
Rename NewWorldRenderingPipeline to EnhancedWorldRenderingPipeline

IMS212:
This naming makes a bit more sense, as it enhances vanilla's pipeline.


coderbot16 (commented):
The original naming is sorta weird (I sorta just named everything 1.17-specific "New") but I don't like this name more. I'll have to think about the naming.

coderbot16:
I'm going to leave the name for now. I'd rather take a chance to start refactoring the 1.16 & 1.17 shader code to remove some duplication and bring them more in line with each other - it's possible that by the end, a lot of the "newshader" stuff will be completely reworked.

PR #793, opened 2021-08-24
Add beta dimension ID map support

IMS212:
Code is unclean, but it works well for the most part.

Only quirk with this is if you have a dimension ID map, you must specify vanilla dimensions yourself.


coderbot16:
One concern I have with this (as well as block.properties and similar) is mod support, as well as datapack support (since datapack dimensions exist).

Maybe there should be a way to have a list of fallbacks / defaults or just a way to match dimensions on something more than just their IDs? Not sure.

Currently, the status quo is that shader packs generally assume everything to be "overworld-ish" unless specified otherwise, would this system support that at least?

coderbot16:
It might not be in the scope of this PR, but it could be cool if we allowed something like "dimension.10=#aether" or similar for popular dimension mods so shader packs wouldn't have to worry about keeping track of the like 5 different aether mods for example.

IMS212:
Mod and datapack dimensions should be fully supported, as this just grabs the ID from the registry key. As for the second one, I can't really answer at this time, as there are many possible ways to do that.

IMS212:
The shader will always fall back to default vanilla worlds if the files don't exist, and if the file does exist but a world isn't specified it'll fall back to the root directory as OptiFine (should.)

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -370,21 +371,17 @@ private static void destroyEverything() {
...
-			if (worldRegistryKey.equals(World.END)) {
-				return DimensionId.END;
-			} else if (worldRegistryKey.equals(World.NETHER)) {
-				return DimensionId.NETHER;
-			} else {
-				return DimensionId.OVERWORLD;
-			}
+			Identifier dimensionId = worldRegistryKey.getValue();
+
+			return getCurrentPack().get().getIdMap().getDimensionIdMap().getOrDefault(dimensionId, 0);
```
IMS212, line 384: Does anyone have a way this could be improved, this is extremely concerning.

IMS212:
image
Proof of this working: Twilight Forest on 1.16 Forge with Complementary Shaders + custom build of Iris + dimension.properties

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -86,21 +83,13 @@ private static ProgramSet loadOverrides(Path root, String subfolder, ShaderPrope
...
-			case END:
-				overrides = end;
-				break;
-			default:
-				throw new IllegalArgumentException("Unknown dimension " + dimension);
+	public ProgramSet getProgramSet(int dimension) {
+		ProgramSet overrides = null;
+
+		try {
+			overrides = loadOverrides(root, "world" + dimension, shaderProperties, this);
```
coderbot16, line 90: We're trying to load all dimension overrides up front in order to avoid disk IO while compiling shaders. That should be possible using the dimensions ID map above.

This is also incorrect, because an unknown or unmatched dimension ID should not map to the world0 override it should map to the root directory.

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
```diff
@@ -18,27 +18,24 @@
...
 
+	private final ShaderProperties shaderProperties;
+	private final Path root;
+
 	public ShaderPack(Path root) throws IOException {
 		// A null path is not allowed.
 		Objects.requireNonNull(root);
 
-		ShaderProperties shaderProperties = loadProperties(root, "shaders.properties")
+		this.shaderProperties = loadProperties(root, "shaders.properties")
```
coderbot16, line 33: ShaderProperties is a temporary intermediate that isn't intended to be stored long-term.

src/main/java/net/coderbot/iris/shaderpack/LegacyIdMap.java
```diff
@@ -56,31 +57,42 @@ public static void addLegacyValues(Object2IntMap<BlockState> blockIdMap) {
...
+	}
+
+	private static void addBlock(Object2IntMap<BlockState> blockIdMap, int id, Block... blocks) {
+		for (Block block : blocks) {
+			block.getStateManager().getStates().forEach(state -> blockIdMap.put(state, id));
 		}
 	}
+
+	private static void addDimension(Object2IntMap<Identifier> dimensionIdMap, int id, Identifier dimension) {
+		dimensionIdMap.defaultReturnValue(0);
```
coderbot16, line 95: This isn't valid - unknown dimensions should be mapped to the root directory, not to world0.

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -70,10 +78,15 @@
 
 		// TODO: Properly override block render layers
 
-		if (blockPropertiesMap == null) {
-			// Fill in with default values...
+		// Fill in with default values...
+		if (blockPropertiesMap == null || blockPropertiesMap.isEmpty()) {
```
coderbot16, line 82: This is invalid - if a shader pack specifies an empty block.properties file, that means that all blocks map to `-1`. Only the absence of a block properties file should be interpreted as the shader pack using legacy mappings.

src/main/java/net/coderbot/iris/shaderpack/IdMap.java
```diff
@@ -70,10 +78,15 @@
...
 			blockPropertiesMap = new Object2IntOpenHashMap<>();
-			LegacyIdMap.addLegacyValues(blockPropertiesMap);
+			LegacyIdMap.addLegacyBlockValues(blockPropertiesMap);
+		}
+
+		if (dimensionIdMap == null || dimensionIdMap.isEmpty()) {
```
coderbot16, line 87: This is invalid - if a shader pack specifies an empty dimension.properties file, that means that all dimensions map to the root directory. Only the absence of a dimension properties file should be interpreted as the shader pack using legacy mappings.

src/main/java/net/coderbot/iris/Iris.java
coderbot16: This is incorrect. An unrecognized dimension should not map to the world0 folder, it should map to the root shaders directory.

coderbot16 (changes requested):
There are a few issues with this implementation that I've pointed out.

Note that there is no integer value that you can use to indicate the absence of a dimension mapping. You will need some other way - whether that is OptionalInt, Integer, or something else.

src/main/java/net/coderbot/iris/shaderpack/ShaderPack.java
IMS212: How would we store all this up front?

IMS212:
Closing as this is extremely outdated, I hope to make a new one soon.

PR #797, opened 2021-08-26
ambientOcclusionLevel constant support

IMS212:
Hopefully fixes some brightness issues with SEUS PTGI, not sure.

Haven't tested this code yet as the only shader that I know uses this is SEUS PTGI which doesn't work on a Macbook.

This will very likely not work with Sodium until we add some helper code to our fork.


coderbot16 (changes requested):
This does not appear to be a correct implementation of ambientOcclusionLevel. I have taken some time to document it here: https://github.com/IrisShaders/ShaderDoc/commit/15964291e37cf8c4a80018d088989a8851b47e9f

IMS212:
This seems to have many issues, the supporting code will need to be reviewed first.

IMS212:
Does my latest rework look better?

IMS212:
Nevermind, found some more major issues.

IMS212:
Issues should be mostly fixed.

src/main/java/net/coderbot/iris/block_rendering/BlockRenderingSettings.java
```diff
@@ -7,6 +7,7 @@
 
 	private boolean reloadRequired;
 	private IdMap idMap;
+	private float ambientOcclusionLevel;
```
coderbot16, line 10: This should always be initialized to 1.0

src/main/java/net/coderbot/iris/mixin/MixinAbstractBlock.java
```diff
@@ -0,0 +1,31 @@
...
+@Mixin(AbstractBlock.class)
+public class MixinAbstractBlock {
+	/**
+	 * @author IMS
+	 * @reason ambientOcclusionLevel support
+	 */
+	@Environment(EnvType.CLIENT)
+	@Deprecated
+	@Overwrite
+	public float getAmbientOcclusionLightLevel(BlockState state, BlockView world, BlockPos pos) {
```
coderbot16, line 22: This doesn't correctly implement the function described in https://github.com/IrisShaders/ShaderDoc/blob/master/const-directives.md#ambient-occlusion-level

Also, if something overrides this function, the resulting value won't be modified properly. It's probably safer to mix in to BlockState / AbstractBlockState instead.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/mixin/MixinAbstractBlockState.java
```diff
@@ -0,0 +1,30 @@
...
+
+	/**
+	 * @author IMS
+	 * @reason ambientOcclusionLevel support
+	 */
+	@Environment(EnvType.CLIENT)
+	@Deprecated
+	@Overwrite
+	public float getAmbientOcclusionLightLevel(BlockView world, BlockPos pos) {
+		float originalValue = this.isFullCube(world, pos) ? 0.2F : 1.0F;
```
coderbot16, line 26: This should really be calling `AbstractBlock#getAmbientOcclusionLightLevel` instead of copying its implementation here.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/mixin/MixinAbstractBlockState.java
IMS212: Isn't that impossible since we're overwriting it?

coderbot16: `AbstractBlock#getAmbientOcclusionLightLevel` is a different method from `AbstractBlockState#getAmbientOcclusionLightLevel`.

IMS212: Done.

src/main/java/net/coderbot/iris/block_rendering/BlockRenderingSettings.java
coderbot16: You might have missed this

IMS212: This is initialized to 1.0 in the constructor?

coderbot16: It's better to give it a default value here instead of just assuming that a constructor will be called and that it will set the value properly

coderbot16 (approved):
I'll fix up the default value of ambientOcclusionLevel myself, this looks good otherwise. Thanks!

PR #823, opened 2021-09-03
Add batched rendering implementation version 2.1

IMS212:


IMS212:
I've disabled the code for the debug information and the sneak to disable, as those were never meant to be in production.

IMS212:
Particles seem to have disappeared, looking into it now.

IMS212:
Fixed.

coderbot16:
Thanks for the PR, but I ended up doing this myself to make sure I didn't miss anything (like implementing WrappingVertexConsumerProvider) and because I wanted a different package layout:

https://github.com/IrisShaders/Iris/compare/e87aacad9584...3fe8f9194de3

PR #825, opened 2021-09-03
Save shadow render distance to configuration file

IMS212:


coderbot16:
Do we want to change this to be an `int` while we have the opportunity?

IMS212:
That would be good, however it doesn't seem like vanilla supports int options without casting.

coderbot16:
We could probably store it as an int wherever we can and cast to/from a double internally as needed.

IMS212:
That works.

IMS212:
I've done this, as well as made some (horrible) code to make the shadow render distance work on all packs, including ones with their own distance.

src/main/java/net/coderbot/iris/config/IrisConfig.java
```diff
@@ -107,6 +108,7 @@ public void load() throws IOException {
 		properties.load(Files.newInputStream(propertiesPath));
 		shaderPackName = properties.getProperty("shaderPack");
 		enableShaders = !"false".equals(properties.getProperty("enableShaders"));
+		IrisVideoSettings.shadowDistance = Integer.parseInt(properties.getProperty("shadowRenderDistance", "0"));
```
coderbot16, line 111: We probably want to add error checking here so that an invalid value gets reset instead of crashing. Also, a small name change. Finally, this should in fact default to 32 - allowing people to effectively disable shadows by setting the distance to zero is desirable.
```suggestion
		IrisVideoSettings.shadowDistance = Integer.parseInt(properties.getProperty("maxShadowRenderDistance", "32"));
```

src/main/java/net/coderbot/iris/gui/option/IrisVideoSettings.java
```diff
@@ -1,22 +1,31 @@
...
+		try {
+			Iris.getIrisConfig().save();
+		} catch (IOException e) {
+			e.printStackTrace();
+		}
 	}, (gameOptions, option) -> {
 		double d = option.get(gameOptions);
 
-		if (d < 0.0) {
+		if (d <= 0.0) {
```
coderbot16, line 28: Allowing people to set the distance to 0.0 to disable shadows is desirable behavior.

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -291,9 +291,11 @@ private Frustum createShadowFrustum() {
 			double distance = halfPlaneLength * renderDistanceMultiplier;
 			String setter = "(set by shader pack)";
 
-			if (renderDistanceMultiplier < 0) {
+			if ((renderDistanceMultiplier < 0 && IrisVideoSettings.shadowDistance != 0) || (IrisVideoSettings.shadowDistance >= 1 && (IrisVideoSettings.shadowDistance * 16) <= halfPlaneLength)) {
 				distance = IrisVideoSettings.shadowDistance * 16;
 				setter = "(set by user)";
+			} else if (renderDistanceMultiplier < 0) {
+				distance = MinecraftClient.getInstance().options.viewDistance * 16;
```
coderbot16, line 298: Similar comments from before about setting the distance to zero.

coderbot16 (changes requested):

IMS212:
All requested suggestions have been made,

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -329,6 +329,10 @@ private Frustum createShadowFrustum() {
 
 	@Override
 	public void renderShadows(WorldRendererAccessor worldRenderer, Camera playerCamera) {
+
+		if (IrisVideoSettings.shadowDistance == 0 && renderDistanceMultiplier < 0) {
+			return;
+		}
```
coderbot16, line 335: This isn't quite right, the shadow map won't be properly cleared by this point AFAIK. We can look more closely into this once I get a chance to look at your ShadowRenderer refactor PR after 1.1.2 is done.

IMS212: Disabled for now.

coderbot16 (approved):
Thanks!

PR #838, opened 2021-09-05
Add option to bundle with sodium to gradle

Razboy20:
Added a gradle task to automatically bundle a build of iris with sodium, getting rid of any need for manually editing and combining the internals of the jars.

After building, you can bundle sodium into the jar through `gradle bundleWithSodium` (and a new files with +sodium suffixes is created).


coderbot16:
Thanks! I merged this in an unconventional way & added a mavenLocal repository declaration.

Razboy20:
Oh yeah, oops I messed up the commits and it should've had the repo in there as well.

PR #839, opened 2021-09-06
Add caching policy to github action, and filter built files

Razboy20:
Should be fine to merge into other branches; also only applies to the build release action.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -5,7 +5,6 @@ on: [ workflow_dispatch ]
 jobs:
   build:
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK 8
@@ -43,10 +42,18 @@ jobs:
         run: ./gradlew publishToMavenLocal
       - name: Build initial Sodium artifacts
         run: cd ./sodium-fabric && ./gradlew publishToMavenLocal && cd ..
+      - name: List contents of maven local repository
+        run: ls -lR /home/runner/.m2/repository
       - name: Build combined Iris + Sodium jar
         run: ./gradlew bundleWithSodium
+      - name: Extract current branch name
+        shell: bash
+        # bash pattern expansion to grab branch name without slashes
+        run: ref="${GITHUB_REF#refs/heads/}" && echo "::set-output name=branch::${ref////-}"
+        id: ref
       - name: Upload build artifacts
         uses: actions/upload-artifact@v2
         with:
-          name: build-artifacts
+          name: iris-artifacts-${{ steps.ref.outputs.branch }}
+          # Filter built files to disregard -sources and -dev, and leave only the minecraft-compatible jars.
           path: build/libs/*[0-9].jar
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -1,4 +1,4 @@
-name: build-release
+name: Build Release
 
 on: [ workflow_dispatch ]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-release.yml
```diff
@@ -12,22 +12,41 @@ jobs:
...
           java-version: 8
+      - name: Cache Iris
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/loom-cache
+            ~/.gradle/wrapper
+            ~/.m2/repository
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
```
coderbot16, line 23: I assume it's intended that gradle.properties isn't included here?

.github/workflows/build-release.yml
```diff
@@ -12,22 +12,41 @@ jobs:
...
+            ./sodium-fabric/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-sodium-${{ hashFiles('./sodium-fabric/**/*.gradle*', './sodium-fabric/**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-sodium-
       - name: Build initial Iris artifacts
         run: ./gradlew publishToMavenLocal
       - name: Build initial Sodium artifacts
         run: cd ./sodium-fabric && ./gradlew publishToMavenLocal && cd ..
-      - name: List contents of maven local repository
-        run: ls -lR /home/runner/.m2/repository
```
coderbot16, line 26: Personally I don't see a huge issue with keeping these, if there are issues with dependencies not being found it's helpful to see the content of mavenlocal.

.github/workflows/build-release.yml
```diff
@@ -12,22 +12,41 @@ jobs:
...
       - name: Build combined Iris + Sodium jar
         run: ./gradlew bundleWithSodium
       - name: Upload build artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v2
         with:
           name: build-artifacts
-          path: build/libs
+          path: build/libs/*[0-9].jar
```
coderbot16, line 52: If possible it would be nice to annotate the intent of this with a comment. I assume this is trying to capture the "-build.*" jars without the "-dev" or "-sources" jars? Maybe a more explicit pattern would be nice (or if there's an explicit way to exclude -dev or -sources with the regex)

.github/workflows/build-release.yml
```diff
@@ -12,22 +12,41 @@ jobs:
...
         run: cd ./sodium-fabric && ./gradlew publishToMavenLocal && cd ..
-      - name: List contents of maven local repository
-        run: ls -lR /home/runner/.m2/repository
       - name: Build combined Iris + Sodium jar
         run: ./gradlew bundleWithSodium
       - name: Upload build artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v2
         with:
           name: build-artifacts
```
coderbot16, line 51: While we're at it, is there any way to make the name of the build artifacts zip be dependent on the branch name? That would be helpful, just curious.

coderbot16 (changes requested):
Thanks! This should be helpful. I have a few comments though.

.github/workflows/build-release.yml
Razboy20: Sadly, the path does not support regex, but rather @action/glob syntax which is weird. I shall add a comment describing it however to make it more clear what the line does.

.github/workflows/build-release.yml
Razboy20: Yes, although it doesn't really matter much tbh. Gradle is pretty smart with its cachings, and custom keys per build.gradle are probably not even needed, but just in case. (This is also the shown example in the cache docs)

Razboy20:
Changed everything that you asked, and also added comments to explain some of the more confusing segments of the action. Artifact names are now `iris-artifacts-<branch name>`, where 1.17/dev would become `iris-artifacts-1.17-dev`

coderbot16 (approved):
Thanks! This should definitely help speed up the actions builds, and the naming of the build artifacts zip should help clear things up as well.

PR #861, opened 2021-09-11
Use bottom Y offset for end portal rendering

### LINKED ISSUE #859: End portal blocks are rendered all the way down with shaders
### What happened?

Unlike staying true to vanilla's (superior) implementation of the fix for https://bugs.mojang.com/browse/MC-3366, the end portal bottom is rendered as if the portal block's hitbox extends all the way to the bottom of the block space (which was the case for 1.16 and below), instead of only half way (which is how it works in 1.17 and above). This not only looks worse than vanilla, but also lets the player clip their camera inside of the end portal texture without being teleported, as demonstrated by this video. https://youtu.be/Z6LDzzPZVis

### Relevant log output

_No response_

Minecraft 1.17.1, Iris 1.1.2

Yes

### Operating System

Windows 10

### What is your GPU?

GTX 1050 Ti

### Java Version

Java 16

### Additional context

_No response_

IMS212:
Fixes #859

There is no bottom Y offset and the block extends to 0.0 on 1.16.


IMS212:
This should be ready to merge.

coderbot16:
Effectively merged with:

- https://github.com/IrisShaders/Iris/commit/12623519a039d62479c65e70e756a0f9307f55af
- https://github.com/IrisShaders/Iris/commit/2767fbf6d70a177cc33ce1848f724a10cb6d1e8b

Thanks!

(I decided to try to apply as much of the diff to trunk as possible first, hence the two commits.)

PR #868, opened 2021-09-13
Add code for registering Iris-exclusive defines and uniforms

IMS212:
This is the same as #771, however without the dragon uniform. (There are some issues with the dragon uniform yet to be sorted out, and I don't want to keep this stalled.)


coderbot16 (approved):
Seems reasonable to me.

PR #871, opened 2021-09-14
Migrate 1.17 to Mojang mappings

IMS212:
This is incredibly unpolished, and almost nothing is renamed at the moment, but it works. Will work on polishing later, feel free to help if you're reading this.


DragonEggBedrockBreaking:
Remember to delete https://github.com/IMS212/Iris/blob/fdfbf71fa569ad23308b3e9156f19574966b59f5/gradle.properties#L6 when it is no longer needed.

IMS212:
This should be ready to merge pending some review.

PR #875, opened 2021-09-18
Add isEyeInCave hardcoded uniform and uniforms for Complementary's biome check

IMS212:
This fixes Complementary 4.2's sky. (This has been band-aid fixed on their side in the newest pre-release, but this is needed for it to properly function.)


IMS212:
I've also added Complementary's biome check, which is needed for rain reflections to properly work.

IMS212:
This should be ready to merge granted the code is cleaned up.

IMS212:
I've added a hacky solution to fix the new if statement to have 0 in water, which involves extending SmoothedFloat.

PR #891, opened 2021-09-24
Remove duplicate draw call

IMS212:
This has literally no reason to be here as far as I can tell, since the batched entity rendering code does this and this just hurts performance.

### DIFF
--- a/src/main/java/net/coderbot/batchedentityrendering/mixin/MixinLevelRenderer.java
+++ b/src/main/java/net/coderbot/batchedentityrendering/mixin/MixinLevelRenderer.java
@@ -21,7 +21,8 @@ import org.spongepowered.asm.mixin.injection.callback.LocalCapture;
  * Tracks whether or not the world is being rendered, and manages grouping
  * with different entities.
  */
-@Mixin(value = LevelRenderer.class)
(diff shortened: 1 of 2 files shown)


coderbot16:
This is needed because otherwise entities will not properly blend with translucent terrain. Squids will be invisible when underwater if you're looking from above water with this PR.

coderbot16:
Oh, wait, I think I see.

IMS212:
The batched entity rendering code has a mixin to the exact same spot running the exact same code.

coderbot16:
Then the concern is figuring out which injection gets applied first - we want the batched entity rendering call to be executed before the Iris call always, but if they have the same injection point and same priority there's no guarantee.

IMS212:
Can we remove the batches entity rendering inject and leave the one here? (The opposite of currently)

coderbot16:
While an option, I would like to try to make the batching code more independent from Iris. Maybe tweaking the injection points / priorities would work.

IMS212:
Tweaking priorities should be fine, we can use a priority one lower

IMS212:
I've changed the batched entity rendering mixin to have a priority of 999, so this should be fine to merge.

coderbot16 (approved):

PR #898, opened 2021-09-27
Add uniforms to fix Complementary's biome check

IMS212:
Not much to explain, this fixes Complementary's aurora and rainbows.


src/main/java/net/coderbot/iris/uniforms/HardcodedCustomUniforms.java
```diff
@@ -50,6 +55,18 @@ private static SmoothedFloat rainStrengthS(FrameUpdateNotifier updateNotifier) {
...
+	private static float getRawPrecipitation() {
+		Biome.Precipitation precipitation = Minecraft.getInstance().level.getBiome(Minecraft.getInstance().player.blockPosition()).getPrecipitation();
+		switch (precipitation.getName()) {
+			case "rain":
+				return 1;
+			case "snow":
+				return 2;
+			default:
+				return 0;
+		}
```
coderbot16, line 67: Please switch on the enum directly instead of converting it to a string first.

src/main/java/net/coderbot/iris/uniforms/HardcodedCustomUniforms.java
```diff
@@ -50,6 +55,18 @@ private static SmoothedFloat rainStrengthS(FrameUpdateNotifier updateNotifier) {
 		return new SmoothedFloat(15, CommonUniforms::getRainStrength, updateNotifier);
 	}
 
+	private static float getRawPrecipitation() {
+		Biome.Precipitation precipitation = Minecraft.getInstance().level.getBiome(Minecraft.getInstance().player.blockPosition()).getPrecipitation();
```
coderbot16, line 59: Might want to use the view entity instead of `Minecraft.getInstance().player`. Also, could getBiome ever return null? It's worth a check.

coderbot16 (changes requested):
I'm slightly concerned at adding more hardcoded custom uniforms especially if some pack authors depend on their presence without properly declaring them in shaders.properties. It might be nice to address that somehow.

src/main/java/net/coderbot/iris/uniforms/HardcodedCustomUniforms.java
IMS212: Fixed, will push shortly after testing.

IMS212:
Pushed.

IMS212:
Everything code-related with this PR should be good to merge.

PR #905, opened 2021-10-01
Added basic support for imageLoad() and imageStore()

BruceKnowsHow:
- Only attached to colorimg0-5 for now
- Textures are probably only allocated if the colortex is used in the shaderpack.
- Image units are inferred based on uniform name.

Here is a shaderpack to test: [test_imageLoadStore.zip](https://cdn.discordapp.com/attachments/774354522361299006/893326695279517746/test_imageLoadStore.zip) Here is a [comparison image](https://cdn.discordapp.com/attachments/254127148913262593/893326955846447134/2021-09-30_19.30.26.png) of what the shaderpack should look like.


IMS212:
Overall looks good, however this needs to be disabled if the machine doesn't support it or it'd probably cause issues.

src/main/java/net/coderbot/iris/gl/texture/PixelFormat.java
```diff
@@ -37,6 +37,73 @@
 		}
 	}
 
+	public static Optional<PixelFormat> fromShaderpackString(String name) {
+		switch(name) {
```
coderbot16, line 41: It feels like these are basically the same values that InternalTextureFormat has. Why not switch on that enum instead of switching on strings? Better yet, why not attach the expected PixelType to the InternalTextureFormat enum?

coderbot16 (changes requested):
Thanks for the PR! Having support for image load store will definitely be helpful. I have a few comments:

- It feels like this PR unnecessarily ties together images & image unit allocation to samplers & sampler unit allocation. From what I can see, as far as OpenGL/GLSL cares, images & samplers aren't connected aside from the fact that they both provide a way of working with texture objects using GLSL. It feels like it would make more sense to create a separate way of attaching images to programs & dynamically allocating image units that is similar to the current sampler system, without having both images & samplers being merged into the *same* system.

- Secondly, did we figure out whether images are influenced by buffer flips? This PR seems to work assuming that they are, however it's possible that they aren't.

- Finally, do we need to disable linear blending when working with integer textures? I heard that could be a problem.

src/main/java/net/coderbot/iris/gl/texture/PixelFormat.java
BruceKnowsHow: I did this. We may want to add PixelType to InternalTextureFormat as well.

BruceKnowsHow:
1. I will look into this, and most create a similar imageUnit class.
2. I'm not sure about this yet. OptiFine doesn't explain how this works, I will test it out and do whatever undefined behavior they do.
3. This might be a problem, I will look into it.

coderbot16:
Hey, is this something you're still interested in working on? Would be nice to see this in Iris at some point. If not, I'd be willing to update these changes to work with the current Iris trunk & 1.17 branches once 1.1.3 is released.

coderbot16:
Thanks for the PR! I've chosen to cherry pick the changes from this PR to make them work with custom textures & the latest trunk branch.

Issue #349, opened 2021-06-09
Unimplemented: "gbuffers_hand" program


EminGT:

coderbot16:
This is because the first-person player hand is not rendered with shaders within the world, but rather as an overlay over the world after world rendering has completed like what vanilla does. This is a low priority for now since in-world shader based hand rendering is nontrivial to implement and has a lot of potential edge cases.

Perhaps https://github.com/tr7zw/FirstPersonModel is a good base for implementing in-world hand rendering.

PR #909, opened 2021-10-02
Hand Rendering

maximumpower55:
closes: #349


Kichura:
https://user-images.githubusercontent.com/68134602/135711324-4621d724-ade1-4fd2-b140-a6e020ff9525.mp4

Seems to be working properly here.

Kichura (approved):

coderbot16:
@Kichura what exactly is the context of the video? The whole world turning black a few seconds in seems like a pretty critical issue

IMS212:
Doesn't seem like you're accounting for non-70 FOV's.

IMS212:
It also seems to clip through blocks

Kichura:
> 
> 
> @Kichura what exactly is the context of the video? The whole world turning black a few seconds in seems like a pretty critical issue

The fading is from continuum itself and i did not make the changes to it, might have to look through the config.

maximumpower55:
Converting this to draft and starting to work on fixing the major clipping issues

maximumpower55:
Fixed Most Major issues

this is some screenshots of it using Complementary v4.2.1

IMS212:
Can you test Sildur's just to make sure

maximumpower55:
Sure one second

IMS212:
It's not, I had the same issue when previously attempting hand rendering which made me stop working on it. Not sure what's up with it.

maximumpower55:
More Debugging for me i guess

maximumpower55:
Looks like its a issue with the way i fixed the clipping issues

Im going to try to find a other way to fix them

maximumpower55:
Cant find a way to fix it

Im going to make this ready for review

If this gets merged

maximumpower55:
Through testing with other shaderpacks this is my collection of issues

Complementary 4.2.1 works with out any issues
Continuum 2.0.4 works with some issues mainly the clouds move with the camera
BSL 8.1.01 works with out any issues
Oceano 3.0.1 is broken
SEUS Renewed 1.0.1 the shadows are broken and the hand is transparent and there is a weird white overlay over the screen
Every Version of Sidurs Vibrant Shaders is broken everything is black but the water

maximumpower55:
@coderbot16 how is this looking for you

coderbot16:
I'm pretty sure clearing the depth buffer is going to cause a lot of chaos. I'll try to figure out how ShadersMod handled hand clipping.

maximumpower55:
ok

maximumpower55:
give all the feedback you have because im a first time contributor to iris

Kichura:
image

Tested complementary here with a gold sword and now it looks worse than before, BSL however doesn't seem to show it here.

maximumpower55:
@Kichura what version of complementary

Kichura:
> 
> 
> @Kichura what version of complementary

v4.2.1 without modifications applied.

maximumpower55:
@Kichura can you screenshot the gold sword in a dark area like a cave

maximumpower55:
@Kichura my gold sword looks fine

Kichura:
> 
> 
> @Kichura can you screenshot the gold sword in a dark area like a cave

image

looks fine in a cave, so might be the sky lighting that caused it.

maximumpower55:
@Kichura looks like its caused by the way i fixed the clipping issues

im waiting on coderbot to figure out how shadersmod handled hand clipping

maximumpower55:
Fixed couple more things

Like the hand rendering when GameRenderer.renderHand was set to false

maximumpower55:
fixed some issues with continuum

src/main/java/net/coderbot/iris/mixin/MixinGameRenderer.java
```diff
@@ -35,7 +35,7 @@
 	// https://github.com/jellysquid3/sodium-fabric/blob/1df506fd39dac56bb410725c245e6e51208ec732/src/main/java/me/jellysquid/mods/sodium/client/render/chunk/shader/ChunkProgram.java#L56
 	@Inject(method = "resetProjectionMatrix", at = @At("HEAD"))
 	private void iris$captureProjectionMatrix(Matrix4f projectionMatrix, CallbackInfo callback) {
-		if(HandRendering.INSTANCE.isRendering()) return;
+		if(HandRenderer.ACTIVE) return; // TODO: This probably should be replaced with something better
```
maximumpower55, line 38: Iris should have a way to stop it from capturing certain things

Justsnoopy30:
I see

maximumpower55:
This is basically help wanted at this point, i can't figure out how ShadersMod handled hand clipping, mostly because the code is very clunky

Justsnoopy30:
If you want to discuss things, you could join the discord and see if someone can help there.

maximumpower55:
Ok did my discord name is maximum btw, i won't be able to talk a lot since busy working on my own mods
usually most active at 5 pm pdt / pt

maximumpower55:
Fixed issues with sidurs ready for review

maximumpower55:
Ready for review

coderbot16:
Initial comments from a quick scan:
- I don't see any implementation of depthtex2 behavior here, which might cause issues with some packs
- gbuffers_hand_water doesn't seem to be implemented
- Hands might end up getting rendered after translucents depending on how the buffering works out
- The code seemingly only weakly uses the gbuffers_hand program for things in the hand - I assume anything with remotely custom rendering will not end up using the hand program

maximumpower55:
@coderbot16 did you not see the commit adding support for hand water

IMS212:
I've added initial support for depthtex2.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,130 @@
...
+		return !(camera.isDetached() 
+			|| !(camera.getEntity() instanceof Player) 
+				|| ((GameRendererAccessor)gameRenderer).getPanoramicMode() 
+					|| Minecraft.getInstance().options.hideGui 
+						|| (camera.getEntity() instanceof LivingEntity && ((LivingEntity)camera.getEntity()).isSleeping()) 
+							|| Minecraft.getInstance().gameMode.getPlayerMode() == GameType.SPECTATOR);
+	}
+
+	public boolean isHandTranslucent(InteractionHand hand) {
+		if (Minecraft.getInstance().player.getItemBySlot(hand == InteractionHand.OFF_HAND ? EquipmentSlot.OFFHAND : EquipmentSlot.MAINHAND).getItem() instanceof BlockItem) {
```
coderbot16, line 52: Should pull this out into

```
Item item = Minecraft.getInstance().player.getItemBySlot(hand == InteractionHand.OFF_HAND ? EquipmentSlot.OFFHAND : EquipmentSlot.MAINHAND).getItem().
```

and then reference it once instead of running that code twice.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,130 @@
...
+			return ItemBlockRenderTypes.getChunkRenderType(((BlockItem) Minecraft.getInstance().player.getItemBySlot(hand == InteractionHand.OFF_HAND ? EquipmentSlot.OFFHAND : EquipmentSlot.MAINHAND).getItem()).getBlock().defaultBlockState()) == RenderType.translucent();
```
coderbot16, line 53: To match OptiFine behavior, I think we need to assume that all hand items are translucent unless proven otherwise (ie, has a known non-translucent render type if it's a block item, and if it's not a block item, assume it's translucent)

src/main/java/net/coderbot/iris/mixin/MixinGameRenderer.java
```diff
@@ -24,4 +34,17 @@
 		Iris.logger.info("GPU: " + GlUtil.getRenderer() + " (Supports OpenGL " + GlUtil.getOpenGLVersion() + ")");
 		Iris.logger.info("OS: " + System.getProperty("os.name"));
 	}
+
+	@Inject(method = "getProjectionMatrix", at = @At(value = "INVOKE", target = "Lcom/mojang/math/Matrix4f;multiply(Lcom/mojang/math/Matrix4f;)V"), locals = LocalCapture.CAPTURE_FAILHARD)
+    private void scaleHandDepth(Camera camera, float f, boolean bl, CallbackInfoReturnable<Matrix4f> cir, PoseStack poseStack) {
+        if (HandRenderer.isActive()) {
+            // This value is taken directly from Shaders Mod.
+            poseStack.scale(1F, 1F, 0.125F);
```
coderbot16, line 42: Would it be possible to move this to HandRenderer as ie. a pre-multiply on the matrix we get from getProjectionMatrix? Combined with other changes, we could remove the HandRenderer.ACTIVE global state.

src/main/java/net/coderbot/iris/mixin/MixinGameRenderer.java
```diff
@@ -24,4 +34,17 @@
...
+        }
+    }
+
+	@Redirect(method = "renderLevel", at = @At(value = "FIELD", target = "Lnet/minecraft/client/renderer/GameRenderer;renderHand:Z"))
+	private boolean disableVanillaHandRendering(GameRenderer gameRenderer) {
+		return false;
```
coderbot16, line 48: We probably don't want to touch vanilla hand rendering when shaders are disabled for any potential mod compatibility issues.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,130 @@
...
+        pose.normal().setIdentity();
+
+		if(Minecraft.getInstance().options.bobView) {
+			((GameRendererAccessor)gameRenderer).invokeBobView(poseStack, tickDelta);
+		}
+	}
+
+	private boolean canRender(Camera camera, GameRenderer gameRenderer) {
+		return !(camera.isDetached() 
+			|| !(camera.getEntity() instanceof Player)
```
coderbot16, line 44: I don't see vanilla explicitly checking for if the camera entity is a player. Is there a reason for that?

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,130 @@
...
+	private void setupGlState(GameRenderer gameRenderer, PoseStack poseStack, float tickDelta, Camera camera) {
+        final PoseStack.Pose pose = poseStack.last();
+
+		// We have a inject in getProjectionMatrix to scale the matrix so the hand doesn't clip through blocks.
+		gameRenderer.resetProjectionMatrix(gameRenderer.getProjectionMatrix(camera, tickDelta, false));
+
+        pose.pose().setIdentity();
+        pose.normal().setIdentity();
+
+		if(Minecraft.getInstance().options.bobView) {
```
coderbot16, line 37: What about bobHurt?

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,130 @@
...
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.BlockItem;
+import net.minecraft.world.level.GameType;
+
+public class HandRenderer {
+	public static final HandRenderer INSTANCE = new HandRenderer();
+
+	private static boolean ACTIVE;
+	public static boolean renderingSolid;
```
coderbot16, line 26: Would rather this be private and have a getter method.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		return false;
+	}
+
+	public boolean isAnyHandTranslucent() {
+		return isHandTranslucent(InteractionHand.MAIN_HAND) || isHandTranslucent(InteractionHand.OFF_HAND);
+	}
+
+	public void renderSolid(RenderBuffers renderBuffers, PoseStack poseStack, float tickDelta, Camera camera, GameRenderer gameRenderer, WorldRenderingPipeline pipeline) {
```
coderbot16, line 63: There's a bit of code duplication between renderSolid and renderTranslucent here.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,130 @@
...
+
+		poseStack.pushPose();
+
+		Minecraft.getInstance().getProfiler().push("iris_hand");
+
+		setupGlState(gameRenderer, poseStack, tickDelta, camera);
+
+		pipeline.pushProgram(GbufferProgram.HAND);
+
+		RenderSystem.disableBlend();
```
coderbot16, line 78: This generally has no effect since render layers always override blend modes. You can probably just remove it.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,130 @@
...
+
+		renderingSolid = true;
+
+		Minecraft.getInstance().getItemInHandRenderer().renderHandsWithItems(tickDelta, poseStack, renderBuffers.bufferSource(), Minecraft.getInstance().player, Minecraft.getInstance().getEntityRenderDispatcher().getPackedLightCoords(camera.getEntity(), tickDelta));
```
coderbot16, line 82: I think we should use a different MultiBufferSource for hand rendering, so we can flush it separately and ensure that all hand content is properly rendered with the hand shader. Otherwise, these buffers might only be rendered by the time HandRenderer becomes inactive (because of how batching and buffering works), meaning that the hand shader might not be used.

The entity ID / block entity ID carefully uses render layer wrapping to accomplish its "refinement" properly, but here we can use a different buffer source, because we really do want to make sure we're flushing it at the right time, separately from the other buffers.

src/main/java/net/coderbot/iris/mixin/rendertype/MixinItemBlockRenderTypes.java
```diff
@@ -0,0 +1,60 @@
...
+import net.coderbot.iris.layer.GbufferProgram;
+import net.coderbot.iris.layer.IrisRenderTypeWrapper;
+import net.coderbot.iris.layer.UseProgramRenderStateShard;
+import net.coderbot.iris.pipeline.HandRenderer;
+import net.minecraft.client.renderer.ItemBlockRenderTypes;
+import net.minecraft.client.renderer.RenderType;
+import net.minecraft.client.renderer.Sheets;
+import net.minecraft.client.renderer.texture.TextureAtlas;
+
+@Mixin(ItemBlockRenderTypes.class)
```
coderbot16, line 20: This entire class could be removed if my comments in MixinRenderType are accurate.

coderbot16 (changes requested):
This is looking much nicer now, good work. I still have some things that I'd like resolved but it's definitely possible this could make it in to the next release.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
IMS212: Should be solved, please review.

src/main/java/net/coderbot/iris/layer/GbufferPrograms.java
```diff
@@ -11,7 +11,7 @@
 	/**
 	 * Uses additional information to choose a more specific (and appropriate) GbufferProgram.
 	 */
-	private static GbufferProgram refine(GbufferProgram program) {
+	public static GbufferProgram refine(GbufferProgram program) {
```
coderbot16, line 14: This should remain private

src/main/java/net/coderbot/iris/mixin/MixinGameRenderer.java
IMS212: I've moved the whole projection matrix code to the hand renderer.

src/main/java/net/coderbot/iris/mixin/rendertype/MixinRenderType.java
```diff
@@ -152,11 +152,7 @@ private static RenderType wrap(RenderType wrapped, GbufferProgram program) {
 	private static void iris$wrapEntityRenderTypes(ResourceLocation texture, CallbackInfoReturnable<RenderType> cir) {
 		RenderType base = cir.getReturnValue();
 
-		if (HandRenderer.isActive()) {
-			cir.setReturnValue(wrap(base, GbufferProgram.HAND));
-		} else {
-			cir.setReturnValue(wrap(base, GbufferProgram.ENTITIES));
-		}
+		cir.setReturnValue(wrap(base, GbufferPrograms.refine(GbufferProgram.ENTITIES)));
```
coderbot16, line 155: This isn't needed - you can just specify ENTITIES, and GbufferPrograms will automatically handle it assuming the global state is correct at the time of actually drawing everything.

```suggestion
		cir.setReturnValue(wrap(base, GbufferProgram.ENTITIES));
```

src/main/java/net/coderbot/iris/mixin/MixinGameRenderer.java
IMS212: Done.

src/main/java/net/coderbot/iris/mixin/GameRendererAccessor.java
```diff
@@ -0,0 +1,21 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Accessor;
+import org.spongepowered.asm.mixin.gen.Invoker;
+
+import net.minecraft.client.renderer.GameRenderer;
+
+@Mixin(GameRenderer.class)
+public interface GameRendererAccessor {
+	@Accessor
+	boolean getRenderHand();
```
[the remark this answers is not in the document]
IMS212: Removed.

src/main/java/net/coderbot/iris/mixin/rendertype/MixinRenderType.java
```diff
@@ -167,11 +163,7 @@ private static RenderType wrap(RenderType wrapped, GbufferProgram program) {
 	private static void iris$wrapEntityRenderTypesZ(ResourceLocation texture, boolean affectsOutline, CallbackInfoReturnable<RenderType> cir) {
 		RenderType base = cir.getReturnValue();
 
-		if (HandRenderer.isActive()) {
-			cir.setReturnValue(wrap(base, GbufferProgram.HAND));
-		} else {
-			cir.setReturnValue(wrap(base, GbufferProgram.ENTITIES));
-		}
+		cir.setReturnValue(wrap(base, GbufferPrograms.refine(GbufferProgram.ENTITIES)));
```
coderbot16, line 166: Same thing here

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -121,10 +127,18 @@ public void renderTranslucent(RenderBuffers renderBuffers, PoseStack poseStack,
 
 		gameRenderer.resetProjectionMatrix(CapturedRenderingState.INSTANCE.getGbufferProjection());
 
+		bufferSource.endBatch();
+
+		bufferSource.resetDrawCalls();
```
coderbot16, line 142: I don't think this does anything particularly important, but it would be good to have it in the other method too to match up.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/mixin/rendertype/MixinItemBlockRenderTypes.java
IMS212: Doesn't seem so, sadly.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -64,8 +65,17 @@ public boolean isAnyHandTranslucent() {
 		return isHandTranslucent(InteractionHand.MAIN_HAND) || isHandTranslucent(InteractionHand.OFF_HAND);
 	}
 
+	public Matrix4f getHandProjectionMatrix(GameRenderer gameRenderer, Camera camera, float tickDelta) {
```
coderbot16, line 68: This isn't really what I meant - I mean that we can take the result of getProjectionMatrix, and construct a new matrix (call it "scale") that just contains the scale call, then call scaleMatrix.multiply(getProjectionMatrix(...)) or something like that.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
coderbot16: That seems to be fair, but I'm pretty sure the behavior is to assume all items that are not BlockItem are translucent, but this code assumes they aren't.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
IMS212: I attempted to do that, but the hand ended up turning invisible due to it not being called before the embedded multiply call of getProjectionMatrix.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
IMS212: We don't want to reset the draw call counter before we've finished everything.

src/main/java/net/coderbot/iris/mixin/rendertype/MixinRenderType.java
IMS212: fixed.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
IMS212: I've just checked again, it does not. It uses the normal hand shader in the case that it is not a block.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
coderbot16: But translucent hand rendering is never guaranteed to be executed.

IMS212: That's true, is it good to reset the draw call count when finalizing level rendering?

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
IMS212: As far as I know no, but it's safer.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
IMS212: Not much we can do about this.

src/main/java/net/coderbot/iris/mixin/rendertype/MixinItemBlockRenderTypes.java
coderbot16: It should be. You might need to return HAND_TRANSLUCENT in GbufferPrograms#refine if the program is TRANSLUCENT_TERRAIN.

coderbot16:
Once MixinItemBlockRenderTypes is addressed this should be ready for merge.

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
```diff
@@ -0,0 +1,137 @@
...
+
+		Minecraft.getInstance().getProfiler().push("iris_hand");
+
+		setupGlState(gameRenderer, camera, poseStack, tickDelta);
+
+		renderingSolid = true;
+
+		Minecraft.getInstance().getItemInHandRenderer().renderHandsWithItems(tickDelta, poseStack, bufferSource, Minecraft.getInstance().player, Minecraft.getInstance().getEntityRenderDispatcher().getPackedLightCoords(camera.getEntity(), tickDelta));
+
+		renderingSolid = false;
```
coderbot16, line 87: This needs to happen after we've flushed the render buffers (ie, the endBatch call)

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/pipeline/HandRenderer.java
IMS212: Fixed.

coderbot16 (approved):
Thanks! The lack of hand rendering is one of the bigger issues affecting Complementary Shaders. I wasn't looking forward to implementing this myself, so it's great that you both took it on. Looking forward to shipping it in the next update!

PR #913, opened 2021-10-03
Delete src/main/java/net/coderbot/iris/mixin/fabulous directory

Just-A-Mango:
It enables compatibility with some other mods, including mine, [Carbon](https://www.curseforge.com/minecraft/mc-mods/carbon)


quat1024:
did you mean with lambdynamiclights? What's Carbon it's 404ing

PR #914, opened 2021-10-03
Delete the NoFabulousSetting Mixin(the class itself and its reference in iris.mixins.json)

Just-A-Mango:
It enables compatibility with some other mods, including mine, [Carbon](https://www.curseforge.com/minecraft/mc-mods/carbon)


IMS212:
This mixin is needed to disable fabulous graphics when using OptiFine shaders, as you cannot use Fabulous graphics with them. This can't be removed, what are you suggesting?

coderbot16:
This Mixin only prevents enabling fabulous graphics with shaders enabled. You can still enable fabulous graphics with shaders disabled.

All this PR does is allow people to enable fabulous graphics with shaders enabled, which makes water invisible in most (if not all) shader packs. Did you even test this PR?

Also, you should know that taking someone else's work and passing it off as your own (as you have done with Carbon, which just copies Bradleyq's mc_vanilla_shaders without giving any credit) is highly frowned upon in the community and in this case is also copyright infringement, which is illegal.

PR #920, opened 2021-10-06
net.coderbot change to net.irisshaders

IMS212:
This would generally help with the view of the project, and not center it around a single point. (Also it just sounds better)


Ruthenic:
can we go ahead and merge this already yall are taking too long

IMS212:
no it won't, that's not how git works. Git tracks each individual line and conflicts based on that, not just if a file is changed.

Ruthenic:
would this not move the file location from `xxx/net/coderbot/yyy` to `xxx/net/irisshaders/yyy` which would break existing prs tho

IMS212:
Git should be smart enough to recognize the move and auto-change, as it tracks renames as well.

IMS212:
I've done a local test and can confirm it works fine in most cases.

IMS212:
Makes sense, I'll mark as draft for now.

coderbot16:
Closing this since this will conflict on literally any change to trunk, and we can easily re-do this when we want to with automated tools. I've created https://github.com/IrisShaders/Iris/issues/1016 to track this.

PR #923, opened 2021-10-06
Update to Fabric Loader 0.12, and use Mixin 0.8.4

IMS212:
This will force 0.12 as the minimum loader, so this should be on hold until 0.12 is considered stable enough for use.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,9 +4,9 @@ org.gradle.jvmargs=-Xmx2G
 # Fabric Properties
 # check these on https://fabricmc.net/versions.html
 minecraft_version=1.16.5
-fabric_version=0.32.0+1.16
+fabric_version=0.40.1+1.16
 
-loader_version=0.11.7
+loader_version=0.12.0
 
 # Mod Properties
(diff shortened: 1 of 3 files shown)


BeansBeefBroccoli:
What did you change in `MixinChunkRebuildTask.java`?

IMS212:
Some Mixin 0.8.4 changes had to be made.

BeansBeefBroccoli:
K

coderbot16 (changes requested):
One of the primary reasons for maintaining 1.16.5 support aside from debugging & potentially better performance is modpack support. Have you checked whether relevant 1.16.5 modpacks are using Fabric Loader 0.12? If not, it might make sense to retain 0.11 compatibility even if we use 0.12 in development.

IMS212:
Just realized this will allow us to get rid of 2 overwrites in MixinBufferVertexConsumer as well due to 0.12.3+ supporting injections into interfaces, so I think we should go ahead. Most 1.16.5 modpacks use 0.12.3+ now.

IMS212:
Will be making a new PR to take advantage of this.

PR #925, opened 2021-10-08
Refactor ShadowRenderer - take two

IMS212:


src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -364,11 +364,119 @@ private Frustum createShadowFrustum() {
...
+
+		renderedShadowBlockEntities = shadowBlockEntities;
+	}
+
 	@Override
 	public void renderShadows(LevelRendererAccessor levelRenderer, Camera playerCamera) {
 		Minecraft client = Minecraft.getInstance();
 
-		levelRenderer.getLevel().getProfiler().popPush("shadows");
+		setProfilerTarget("shadows");
```
coderbot16, line 479: I don't see the need for this particular change. It would make more sense to store the profiler at the head of the method or something to avoid repetition though,

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
IMS212: Will do

IMS212: done

coderbot16:
I'll investigate including this for 1.1.4 / 1.2.0 but there's too little time to test this PR for the next release, so I removed the "next release" label.

coderbot16 (approved):
Thanks!

PR #943, opened 2021-10-25
Tweak view distance based on fog in 1.16

IMS212:
In 1.17, we should hardcode this to 0.95.


coderbot16:
Where are you getting this 0.95 value from? And have you checked the impact that this has on water reflections?

coderbot16:
Oh, I see, had to look at my own code comment.

PR #956, opened 2021-11-03
Comment out "test", which now means Iris can be built.

BeansBeefBroccoli:
/*
test {
	// Use JUnit 5 Platform for tests
	useJUnitPlatform()
}
*/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,6 +108,7 @@ sourceSets {
 		}
 	}
 
+/*
 	test {
 		java {
 			// headers / API stubs are only available at compilation time.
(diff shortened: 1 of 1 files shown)


IMS212:
This isn't a proper fix, a proper fix would be actually fixing what makes the tests fail.

BeansBeefBroccoli:
Ok, I just wanted to do this because the feature wasn't fixed in the next, like, 3 commits, and I wanted to fix this for others until the feature gets properly introduced.

BeansBeefBroccoli:
I have no idea how to code graphics with MC, if I did, I would try to actually help with the feature.

IMS212:
Closed in favor of a new PR I'm making which correctly fixes this

BeansBeefBroccoli:
Ok, thank you! :+1:

PR #957, opened 2021-11-03
Temporary JUnit fix

IMS212:
This is a bad hack, but it has the tests correctly working.


IMS212:
Having an boolean that states if we are currently testing could be useful in the future as well, though it may be better to have it go somewhere else.

coderbot16:
This isn't really the approach I'd like to take - I'd rather make it so that our shader pack loader doesn't reference Minecraft classes in the first place. But I've been waiting until I have a good chance to refactor IdMap to not present Minecraft classes in its API.

coderbot16:
Created a note for this so I remember to fix the underlying issue in 1.1.4: https://github.com/IrisShaders/Iris/issues/1017

PR #967, opened 2021-11-11
Prefix vanilla attributes with iris_ on 1.17

IMS212:
This matches Optifine behavior, and fixes a few incompatibilities from shaders using the same attribute names.


coderbot16:
I was waiting until we got the new shader patcher to do something like this, because this won't work with terrain shaders when using Sodium. I was thinking of prefixing the attributes with `iris_` to prevent unintentional conflicts, and shaders expecting `va` will be handled with the new patching setup.

IMS212:
That makes sense, should I change this to prefix with iris_? Right now the current names apparantly cause new packs like NostalgiaVX to break due to conflicts.

coderbot16:
For now, an `iris_` prefix makes the most sense.

src/main/java/net/coderbot/iris/pipeline/newshader/OptifineCorePatcher.java
```diff
@@ -0,0 +1,37 @@
+package net.coderbot.iris.pipeline.newshader;
+
+import net.coderbot.iris.gl.blending.AlphaTest;
+import net.coderbot.iris.shaderpack.transform.StringTransformations;
+import net.coderbot.iris.shaderpack.transform.Transformations;
+
+public class OptifineCorePatcher {
```
coderbot16, line 7: Would like for this to be pulled out of this PR, ideally we'd hold off on this until we implement it in the rewritten patching system based on glsl-transformer.

IMS212: 👍

IMS212: Done.

src/main/java/net/coderbot/iris/mixin/MixinShaderInstance.java
```diff
@@ -74,4 +76,13 @@
 
 		logger.warn(message, arg1, arg2);
 	}
+
+	@Redirect(method = "<init>", at = @At(value = "INVOKE", target = "Lcom/mojang/blaze3d/shaders/Uniform;glBindAttribLocation(IILjava/lang/CharSequence;)V"))
+	public void iris$redirectBindAttributeLocation(int i, int j, CharSequence charSequence) {
+		if (((Object) this) instanceof ExtendedShader && Arrays.asList("Position", "Color", "Normal", "UV0", "UV1", "UV2").contains(charSequence)) {
```
coderbot16, line 82: List#contains is inefficient, not to mention constructing a new time every time we call bind. Best to create a HashSet / ImmutableSet once & call contains using that instead.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/mixin/MixinShaderInstance.java
IMS212: Done.

src/main/java/net/coderbot/iris/mixin/MixinShaderInstance.java
```diff
@@ -15,15 +16,16 @@
...
-import java.util.Arrays;
 import java.util.Objects;
 
 @Mixin(ShaderInstance.class)
 public class MixinShaderInstance {
 	@Unique
 	private String lastSamplerName;
 
+	@Unique
+	private final ImmutableSet<String> attributeList = ImmutableSet.of("Position", "Color", "Normal", "UV0", "UV1", "UV2");
```
coderbot16, line 27: probably best to do `static` & `ATTRIBUTE_LIST`

coderbot16 (commented):
Thanks! I'll fix that myself.

PR #981, opened 2021-11-16
Detach shaders from program after linking

maximumpower55:
A simple fix to a small memory leak

This is 1.16 only because 1.17 iris uses the vanilla shader system witch fixes this

For more info look at this sodium pr [link](https://github.com/CaffeineMC/sodium-fabric/pull/971)


IMS212:
Looks good to me.

IMS212 (approved):
Looks good to me.

coderbot16 (approved):
This does not seem to cause an actual memory leak, but it is still a good change regardless. Thanks!

Issue #983, opened 2021-11-17
Remaining things to implement with forkless Sodium compatibility


coderbot16:
- 1.16.5 issues
    - [X] Make 1.16.5 link to https://www.curseforge.com/minecraft/mc-mods/sodium/files/3488820 instead.
- 1.17.1+ issues
    - [x] MixinWorldRenderer: Do not cancel skybox rendering if shaders are enabled
    - [x] #990
    - [x] ~~Check if https://github.com/IrisShaders/Iris/issues/792 was fixed~~ Edit: not fixed
    - [x] Figure out how to deal with the Shadow remapping issue in dev other than by disabling `mixin.features.texture_tracking` and `mixin.features.texture_updates`
- Common issues
    - [x] ~~Loosen Sodium version locks with SemVer~~ Sodium doesn't use SemVer

IMS212:
I have a PR #986 to fix the alpha cutoff.

coderbot16:
- [ ] Lock the Sodium version to an exact version to stop people from running with the wrong Sodium version, because 0.3 is too loose.

coderbot16:
https://github.com/IrisShaders/Iris/issues/792 is confirmed to *not* be fixed.

IMS212:
Everything here should be fixed as of now,

PR #986, opened 2021-11-17
Fix alpha cutoff

IMS212:
Turns out the only render pass with a 0.5 alpha is cutoutMipped anyway, so we can just use an if statement.


Justsnoopy30 (approved):
LGTM

Justsnoopy30:
Additionally, I've tested the PR and it appears to work fine.

IMS212:
Fixes part of #983

coderbot16 (approved):
Thanks!

PR #988, opened 2021-11-17
Support for getNightVisionScale injections

Mourdraug:
These changes allow nightvision to work for mods which achieve it by directly injecting code into getNightVisionScale (For example [Origins](https://github.com/apace100/apoli/blob/320b0ef547fbbf703de7154f60909d30366f6500/src/main/java/io/github/apace100/apoli/mixin/GameRendererMixin.java#L153)) 
This also means that non binary nightVision values are now possible.

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/MixinGameRenderer.java
+++ b/src/main/java/net/coderbot/iris/mixin/MixinGameRenderer.java
@@ -34,4 +34,11 @@ public class MixinGameRenderer {
 	private boolean disableVanillaHandRendering(GameRenderer gameRenderer) {
 		return !Iris.getCurrentPack().isPresent();
 	}
+	
+	@Inject(method = "getNightVisionScale", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/effect/MobEffectInstance;getDuration()I"), cancellable = true)
+	private static void iris$safecheckNightvisionStrength(LivingEntity livingEntity, float f, CallbackInfoReturnable<Float> cir){
+		if(livingEntity.getEffect(MobEffects.NIGHT_VISION) == null){
+			cir.setReturnValue(0.0f);
+		}
+	}
 }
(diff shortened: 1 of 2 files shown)


AlsoSylv (approved):
Everything looks good to me

src/main/java/net/coderbot/iris/mixin/MixinGameRenderer.java
```diff
@@ -34,4 +34,11 @@
 	private boolean disableVanillaHandRendering(GameRenderer gameRenderer) {
 		return !Iris.getCurrentPack().isPresent();
 	}
+	
+	@Inject(method = "getNightVisionScale", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/effect/MobEffectInstance;getDuration()I"), cancellable = true)
+	private static void iris$safecheckNightvisionStrength(LivingEntity livingEntity, float f, CallbackInfoReturnable<Float> cir){
+		if(livingEntity.getEffect(MobEffects.NIGHT_VISION) == null){
```
coderbot16, line 40: This injection could apply before Origins, and might actually break the effect in the mod completely.

coderbot16 (changes requested):
This is a nice start, however, I think we could do better with a different injection. It seems like Origins 1.16.x injects into LightTexture (mojmap) / LightmapTextureManager (yarn), so we could instead maybe use a local variable capture in LightTexture to get the night vision value and store that somewhere, then pass that value to shaders in CommonUniforms.

Obviously local variable captures are fragile, and the injection has to be a bit precise, but I think they might be better here. Alternatively, we can have a separate GameRenderer mixin with a different priority that would guarantee that it applies before the mixin from Origins.

coderbot16:
Thanks for the PR! Looks like my initial comment was incorrect, upon another review this code looks good & should be a nice workaround for Origins on 1.16.

PR #997, opened 2021-11-21
Semantic versioning support for Sodium

IMS212:


IMS212:
Fixes part of #983

coderbot16:
Sodium doesn't use SemVer in a way that helps us, so we can't do this, unfortunately. All of the builds on 0.4-experiments are labeled as builds of 0.3.2 version-wise, as well as the breaking changes (for Iris) on the Sodium 1.17.x/dev. I've updated the original issue accordingly.

PR #1000, opened 2021-11-23
Sodium 0.3.3 compatibility

IMS212:
This should preferably be merged as soon as 0.3.3 is released.


IMS212:
Just noticed this is issue #1000!

PR #1002, opened 2021-11-23
Remove unnecessary info about the Sodium fork

batu-0:
This is my first pull request to a public project, so I just wanted to start with something really basic that I noticed.
Iris no longer requires the Sodium fork to run together with Sodium, so we should update the readme accordingly.
This really is just a minor change. I didn't know what to replace the removed part with, so please request changes if needed.


BeansBeefBroccoli:
LGTM :+1:

coderbot16:
Thanks! I might write some text about how a Sodium fork was previously required but is no longer needed.

PR #1004, opened 2021-11-23
Separate shader patching from SodiumTerrainPipeline init, use vertex type info in patching

IMS212:


IMS212:
Merging as this has been tested for over 3 days as part of a public test build with no issues.

PR #1010, opened 2021-11-28
Custom blending mode support

IMS212:
The new refractor should make this ready to merge.


coderbot16:
Thanks! I've applied a few fixes myself and this should be good to go now.

PR #1013, opened 2021-11-29
Update Korean translation

craftingmod:
Korean language was translated with google translate and quality is bad.

Updated to recent json with some translation fix.

1. Korean calls Iris to `아이리스` so I translated Iris to `아이리스` except title. 
2021-11-30_03-50-59_msedge
2. Because I need to search English mod name (Sodium = `소듐`, Iris = `아이리스` for example), I also put Sodium's English name for translation.
3. Refresh and Reload are similar so Refresh is translated to `새로고침` and Reload is translated to `다시 불러오다`. toggle is translated to `ON/OFF 전환(switch)` or `%s로 변경` (set to %s) .
4. I know the most correct translation of `shaders` depended on Korean English conversion guide is `셰이더` but everyone is using `쉐이더`.. So I changed to `쉐이더` for compactivity.
5. Else, my opinion 🤔


craftingmod:
`앤`(and) -> &

coderbot16:
Thanks!

PR #1040, opened 2021-12-01
Add automatic release publishing

NoComment1105:
This alters the release workflow to automatically push builds to Curseforge Modrinth and Github. This will need to be copied over to 1.17 and 1.18 meaning 3 different workflows will need to be run at each release. The code is entirely copiable and not version bound

I know you wanterd  a seperate gradle file but pain :)

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -18,11 +18,43 @@ jobs:
       - name: Cache Gradle packages
         uses: actions/cache@v2
         with:
-          path: ~/.gradle/caches
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/loom-cache
+            ~/.gradle/wrapper
+            ~/.m2/repository
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
           restore-keys: ${{ runner.os }}-gradle
-      - name: Upload artifacts to Curse, Modrinth and Github
-        run: ./gradlew build curseforge github modrinth
+
+      - run: chmod +x gradlew
+      - name: Build Release
+        run: ./gradlew build --stacktrace
+      - name: Upload release Builds to Github, Curseforge and Modrinth
+        run: ./gradlew github curseforge modrinth --stacktrace
+        env:
+          CURSEFORGE_TOKEN: ${{ secrets.CURSEFORGE_TOKEN }}
+          MODRINTH_TOKEN: ${{ secrets.MODRINTH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Switch to 1.17
+        run: git checkout 1.17
+
+      - name: Build 1.17
+        run: ./gradlew build --stacktrace
+      - name: Upload 1.17 Builds to Github, Curseforge and Modrinth
+        run: ./gradlew github curseforge modrinth --stacktrace
+        env:
+          CURSEFORGE_TOKEN: ${{ secrets.CURSEFORGE_TOKEN }}
+          MODRINTH_TOKEN: ${{ secrets.MODRINTH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Switch to 1.18
+        run: git checkout 1.18
+
+      - name: Build 1.18
+        run: ./gradlew build --stacktrace
+      - name: Upload 1.18 Builds to Github, Curseforge and Modrinth
+        run: ./gradlew github curseforge modrinth --stacktrace
         env:
           CURSEFORGE_TOKEN: ${{ secrets.CURSEFORGE_TOKEN }}
           MODRINTH_TOKEN: ${{ secrets.MODRINTH_TOKEN }}
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -26,7 +26,7 @@ jobs:
           path: ~/.gradle/caches
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
-      - name: Upload artifacts to Curse/Modrinth
+      - name: Upload artifacts to Curse, Modrinth and Github
         run: ./gradlew build curseforge github modrinth
         env:
           CURSEFORGE_TOKEN: ${{ secrets.CURSEFORGE_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build-release.yml
```diff
@@ -1,37 +1,34 @@
...
       - uses: actions/checkout@v2
-      - name: Set up JDK 8
-        uses: actions/setup-java@v1
         with:
-          java-version: 8
-      - name: Cache Iris
-        uses: actions/cache@v2
+          fetch-depth: 0
+      - uses: FabricMC/fabric-action-scripts@v1
+        id : changelog
```
coderbot16, line 19: Is this needed? We usually manually write changelogs anyway, because that always gives us a nicer and more user-friendly result.

.github/workflows/build-release.yml
```diff
@@ -1,37 +1,34 @@
...
+          context: changelog
+          workflow_id: build-release.yml
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
         with:
-          name: iris-artifacts-${{ steps.ref.outputs.branch }}
-          # Filter built files to disregard -sources and -dev, and leave only the minecraft-compatible jars.
-          path: build/libs/*[0-9].jar
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
```
coderbot16, line 27: Previously, we used:

```
key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
```

I assume we still want to use gradle-wrapper.properties as a key in case gradle versions are significant.

build.gradle
```diff
@@ -237,3 +247,98 @@ def getVersionMetadata() {
...
+	}
+
+	doLast {
+		def github = GitHub.connectUsingOAuth(ENV.GITHUB_TOKEN as String)
+		def repository = github.getRepository(ENV.GITHUB_REPOSITORY)
+
+		def releaseBuilder = new GHReleaseBuilder(repository, version as String)
+		releaseBuilder.name("Iris $project.mod_version for Minecraft $project.minecraft_version")
+		releaseBuilder.body("https://github.com/NoComment1105/Iris/tree/trunk/docs/changelogs/${project.mod_version}summary.md")
+		releaseBuilder.commitish(getBranch())
```
coderbot16, line 338: Is this needed? Seems like we're converting from commit ID to branch, would it be more direct to just pass the commit ID?

build.gradle
```diff
@@ -237,3 +247,98 @@ def getVersionMetadata() {
...
+		ENV.GITHUB_TOKEN
+	}
+
+	doLast {
+		def github = GitHub.connectUsingOAuth(ENV.GITHUB_TOKEN as String)
+		def repository = github.getRepository(ENV.GITHUB_REPOSITORY)
+
+		def releaseBuilder = new GHReleaseBuilder(repository, version as String)
+		releaseBuilder.name("Iris $project.mod_version for Minecraft $project.minecraft_version")
+		releaseBuilder.body("https://github.com/NoComment1105/Iris/tree/trunk/docs/changelogs/${project.mod_version}summary.md")
```
coderbot16, line 337: Wrong repo :P

Also, could this just paste the body of the summary in?

build.gradle
```diff
@@ -237,3 +247,98 @@ def getVersionMetadata() {
...
+		releaseBuilder.commitish(getBranch())
+		releaseBuilder.prerelease(false)
+
+		def ghRelease = releaseBuilder.create()
+		ghRelease.uploadAsset(file("${project.buildDir}/libs/${project.archivesBaseName}-${project.mod_version}+build.${ENV.GITHUB_RUN_NUMBER}.jar"), "application/java-archive");
```
coderbot16, line 342: We're not interested in including the build number for official full releases.

coderbot16 (changes requested):
Thanks! I have a few comments. In addition, there's two things I would like:

- Having the release automation code in a separate `.gradle` file that can be easily disabled is important to me to avoid pulling in these plugins unnecessarily in a development environement (also to avoid fragility).
- It would be nice to have some sort of "dry run" option that would simulate a lot of the steps but not actually upload the JARs to GH / Curse / Modrinth, and instead just let us download them, as a way to verify that the JARs aren't bad before shipping them.

.github/workflows/build-release.yml
NoComment1105: Not sure, It got a bit funny when I didn't have it, but I'll try again later today

.github/workflows/build-release.yml
NoComment1105: Ah yikes, I missed adding that back in

build.gradle
NoComment1105: I did consider just simply uploading the file from summary but my time got tight, and previous releases just provided a link, and as such I assumed a link would be fine

build.gradle
NoComment1105: Trouble is the file that gets automatically generated comes with a build number, which will happen for all of these uploads, since github is running the build. If you really don't want it I'll have to rethink this

build.gradle
NoComment1105: Possibly yes, this is left from when I was trying to get it to loop through all 3 branches to upload, however I'm going to change the approach for it.

.github/workflows/build-release.yml
NoComment1105: Solves

.github/workflows/build-release.yml
NoComment1105: Removed and **seems** to be working

build.gradle
NoComment1105: Turns out it's not needed at all as by default it gets the branch name

build.gradle
NoComment1105: Fixed this.

build.gradle
NoComment1105: I was unable to find a way to upload a file, since one code completion wouldn't work for me, and two `releaseBuilder.body` only takes a string

NoComment1105:
Just need to update the summary of this pr

This pr adds automatic uploading of releases to Github Modrinth and CurseForge via 3 gradle tasks and a modified build-release.yml. the actual publishing code will follow branch specific information, and is controlled in the release file. After building and releasing trunk, the script automatically checkouts to 1.17 and 1.18 before repeating. At the moment the automatic Changelog is a link to the summary in the docs section. However that can be updated to send the contents of the file in the Modrinth and CurseForge release scripts. I am unsure about the Github script, since the builder only takes string and refuses to take a file, plus auto complete being a 🦀 for me

PR #1046, opened 2021-12-04
Update zh_cn translations

klkq:
Added entries.
Polished translations.
Using "。" instead of "." to end a sentence.


coderbot16 (approved):
Thanks!

Issue #129, opened 2021-02-26
Allow shaderpacks to disable the water overlay


coderbot16:
https://github.com/sp614x/optifine/blob/ac35062118c29593b7540f492a3af6b0a7da3391/OptiFineDoc/doc/shaders.properties#L44

PR #1056, opened 2021-12-06
Implement the underwater overlay property

maximumpower55:
Closes: #129

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/MixinScreenEffectRenderer.java
+++ b/src/main/java/net/coderbot/iris/mixin/MixinScreenEffectRenderer.java
@@ -10,7 +10,7 @@ import org.spongepowered.asm.mixin.injection.Inject;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 
 import net.coderbot.iris.Iris;
-import net.coderbot.iris.shaderpack.ShaderPack;
+import net.coderbot.iris.pipeline.WorldRenderingPipeline;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.ScreenEffectRenderer;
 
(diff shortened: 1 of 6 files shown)


coderbot16 (changes requested):
Would it be possible to control this with a method on WorldRenderingPipeline instead?

coderbot16:
This is an example of how that might look:

https://github.com/IrisShaders/Iris/blob/528488f7c050d077edd76b0b7974aa14fb5ba9e2/src/main/java/net/coderbot/iris/mixin/MixinEntityRenderDispatcher.java

maximumpower55:
> This is an example of how that might look:
> 
> https://github.com/IrisShaders/Iris/blob/528488f7c050d077edd76b0b7974aa14fb5ba9e2/src/main/java/net/coderbot/iris/mixin/MixinEntityRenderDispatcher.java

Will look into this tomorrow

src/main/java/net/coderbot/iris/mixin/MixinScreenEffectRenderer.java
```diff
@@ -0,0 +1,27 @@
...
+@Mixin(ScreenEffectRenderer.class)
+public abstract class MixinScreenEffectRenderer {
+	@Inject(method = "renderWater", at = @At(value = "HEAD"), cancellable = true)
+	private static void iris$disableUnderWaterOverlayRendering(Minecraft minecraft, PoseStack poseStack, CallbackInfo ci) {
+		WorldRenderingPipeline pipeline = Iris.getPipelineManager().getPipelineNullable();
+
+		if (pipeline != null && !pipeline.shouldRenderUnderwaterOverlay()) {
+			ci.cancel();
+		}
+    }
```
coderbot16, line 26: wrong indentation?

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/mixin/MixinScreenEffectRenderer.java
maximumpower55: Code_VyMWmCeXyD

i think thats just git

maximumpower55: Fixed

coderbot16 (approved):
Thanks!

PR #1061, opened 2021-12-06
Modernize ShadowRenderer for 1.17

IMS212:
Huge changes, do not merge yet.

### DIFF
--- a/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
+++ b/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
@@ -12,18 +12,11 @@ import net.coderbot.batchedentityrendering.impl.BatchingDebugMessageHelper;
 import net.coderbot.batchedentityrendering.impl.DrawCallTrackingRenderBuffers;
 import net.coderbot.batchedentityrendering.impl.RenderBuffersExt;
 import net.coderbot.iris.Iris;
-import net.coderbot.iris.gl.blending.BlendModeOverride;
-import net.coderbot.iris.gl.program.Program;
-import net.coderbot.iris.gl.program.ProgramBuilder;
-import net.coderbot.iris.gl.program.ProgramSamplers;
 import net.coderbot.iris.gl.texture.InternalTextureFormat;
 import net.coderbot.iris.gl.framebuffer.GlFramebuffer;
 import net.coderbot.iris.gui.option.IrisVideoSettings;
-import net.coderbot.iris.layer.GbufferProgram;
 import net.coderbot.iris.mixin.LevelRendererAccessor;
-import net.coderbot.iris.pipeline.newshader.FogMode;
 import net.coderbot.iris.rendertarget.RenderTargets;
-import net.coderbot.iris.samplers.IrisSamplers;
 import net.coderbot.iris.shaderpack.*;
 import net.coderbot.iris.shadow.ShadowMatrices;
 import net.coderbot.iris.shadows.CullingDataCache;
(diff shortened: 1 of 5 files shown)


src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -148,48 +117,33 @@ public ShadowRenderer(WorldRenderingPipeline pipeline, ProgramSource shadow, Pac
...
-		BlendModeOverride blendModeOverride;
 
 		if (shadow != null) {
-			//this.shadowProgram = createProgram(shadow, directives, flipped);
-
-			blendModeOverride = shadow.getDirectives().getBlendModeOverride();
-
-			if (blendModeOverride == null) {
-				blendModeOverride = BlendModeOverride.OFF;
-			}
```
coderbot16, line 162: Don't forget to replicate this behavior

IMS212: No need, this is handled automatically as part of ExtendedShader. Or at least it seems so?

coderbot16: No, the behavior where the blend mode override defaults to OFF instead of no override for shadow programs is not handled by ExtendedShader.

IMS212: Done.

src/main/java/net/coderbot/iris/pipeline/newshader/ExtendedShader.java
```diff
@@ -73,6 +73,8 @@ public void apply() {
 
 		if (this.blendModeOverride != null) {
 			this.blendModeOverride.apply();
+		} else if (getName().contains("shadow")) {
+			BlendModeOverride.OFF.apply();
```
coderbot16, line 77: Would it be possible to handle this where we don't have to check the name of the shader program? If there's a nice way to avoid checking the name, we should do that.

IMS212: Doesn't seem so, unless we want to pass through arguments to say it's a shadow shader.

src/main/java/net/coderbot/iris/shadows/ShadowRenderingState.java
```diff
@@ -9,6 +10,6 @@ public static boolean areShadowsCurrentlyBeingRendered() {
 	}
 
 	public static Matrix4f getShadowOrthoMatrix() {
```
coderbot16, line 12: Is this method even used now?

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -484,16 +380,15 @@ public void renderShadows(LevelRendererAccessor levelRenderer, Camera playerCame
...
 
-		Matrix4f projectionMatrix = new Matrix4f();
-		projectionMatrix.load(projMatBuf);
+		Matrix4f projMatrix = new Matrix4f();
+		((Matrix4fAccess) (Object) projMatrix).copyFromArray(orthoMatrix);
 
-		Matrix4f previousProjectionMatrix = RenderSystem.getProjectionMatrix();
-		RenderSystem.setProjectionMatrix(projectionMatrix);
+		// Load projection matrix into RenderSystem and backup the old one
+		RenderSystem.backupProjectionMatrix();
```
coderbot16, line 390: We should probably just save the projection matrix in our own variable in case something else tries to call backupProjectionMatrix in shadow rendering

src/main/java/net/coderbot/iris/pipeline/newshader/ExtendedShader.java
coderbot16: I'd like to find a way to address this before merging.

coderbot16 (changes requested):
This is a nice improvement, but I'd like to find a way to address the blend mode overriding issue before merging. I'll think of something.

PR #1063, opened 2021-12-07
Use default vertex format with shaders disabled (Sodium compatibility)

IMS212:


src/sodiumCompatibility/java/net/coderbot/iris/compat/sodium/impl/vertex_format/xhfp/XHFPModelVertexBufferWriterNio.java
```diff
@@ -16,7 +19,7 @@
     private MaterialIdHolder idHolder;
 
     public XHFPModelVertexBufferWriterNio(VertexBufferView backingBuffer) {
-        super(backingBuffer, IrisModelVertexFormats.MODEL_VERTEX_XHFP);
+        super(backingBuffer, Iris.isPackActive() ? IrisModelVertexFormats.MODEL_VERTEX_XHFP : DefaultModelVertexFormats.MODEL_VERTEX_HFP);
```
coderbot16, line 22: This probably doesn't need to be changed - this class should only ever be used if the extended vertex format is being actively used.

src/sodiumCompatibility/java/net/coderbot/iris/compat/sodium/mixin/vertex_format/MixinSodiumWorldRenderer.java
```diff
@@ -21,6 +24,6 @@
                             ")Lme/jellysquid/mods/sodium/client/render/chunk/ChunkRenderBackend;"))
     private ChunkVertexType iris$overrideVertexType(ChunkVertexType vertexType) {
         // TODO: Don't use the extended vertex format when shaders are disabled.
-        return IrisModelVertexFormats.MODEL_VERTEX_XHFP;
+        return Iris.isPackActive() ? IrisModelVertexFormats.MODEL_VERTEX_XHFP : DefaultModelVertexFormats.MODEL_VERTEX_HFP;
```
coderbot16, line 27: We can just return the `vertexType` if the pack is inactive here. Also, you can probably remove that TODO.

coderbot16 (changes requested):

IMS212:
Fixed both reviews.

src/sodiumCompatibility/java/net/coderbot/iris/compat/sodium/mixin/vertex_format/MixinSodiumWorldRenderer.java
```diff
@@ -21,6 +24,6 @@
                             ")Lme/jellysquid/mods/sodium/client/render/chunk/ChunkRenderBackend;"))
     private ChunkVertexType iris$overrideVertexType(ChunkVertexType vertexType) {
         // TODO: Don't use the extended vertex format when shaders are disabled.
```
coderbot16, line 26: You can probably remove this comment since you're making the change that resolves it.

IMS212: Done

coderbot16 (approved):
Thanks! If you port this to the 1.17 branch you should PR that too just since this is a sensitive area. We probably want to get this code decently tested before releasing.

PR #1079, opened 2021-12-12
Separate Sodium, Iris, and Minecraft versions in issue templates

tempest15:
Given that Iris no longer uses bundled Sodium it's important to be able to easily know what Sodium version a user has without going into their log file. This pull request separates the fields for Iris, Sodium, and Minecraft versions. I've seen several recent issues that would've been resolved faster if the information were presented in this format.


coderbot16 (approved):
Looks good to me, thanks.

PR #1090, opened 2021-12-13
Update Release checklist

NoComment1105:
Does what it says on the tin. Simply to match the new action system


NoComment1105:
Ok commit history is screwed excuse me

PR #1096, opened 2021-12-15
Update to Fabric Loader 0.12.12, use Mixin 0.8.4

IMS212:
This allows us to get rid of 2 overwrites in favor of ModifyVariable.

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -45,7 +45,7 @@
   "accessWidener" : "iris.accesswidener",
 
   "depends": {
-    "fabricloader": ">=0.7.4",
+    "fabricloader": ">=0.12.3",
     "minecraft": "1.16.5",
     "fabric-resource-loader-v0": ">=0.3.2",
     "fabric-key-binding-api-v1": ">=1.0.1"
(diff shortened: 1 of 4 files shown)


src/main/resources/fabric.mod.json
```diff
@@ -45,7 +45,7 @@
   "accessWidener" : "iris.accesswidener",
 
   "depends": {
-    "fabricloader": ">=0.7.4",
+    "fabricloader": ">=0.12.3",
```
DragonEggBedrockBreaking, line 48:
```suggestion
    "fabricloader": ">=0.12.32”,
```

DragonEggBedrockBreaking: Idk why GitHub formatted it so weirdly…

IMS212: Are you sure 0.12.2 works for this?

DragonEggBedrockBreaking: Oops

coderbot16 (approved):
Thanks. I'd like to replace the relevant Mixin entirely at some point but newer Mixin is nice. If the new Fabric Loader version requirement turns out to be an issue for modpacks we can always downgrade later.

PR #1101, opened 2021-12-16
Implement the rain.depth property

maximumpower55:


src/main/java/net/coderbot/iris/mixin/MixinLevelRenderer.java
```diff
@@ -187,6 +187,15 @@
 		pipeline.pushProgram(GbufferProgram.WEATHER);
 	}
 
+	@Inject(method = RENDER_WEATHER, at = @At(value = "INVOKE", target = "Lcom/mojang/blaze3d/systems/RenderSystem;depthMask(Z)V", shift = At.Shift.AFTER))
+	private void iris$writeRainAndSnowToDepthBuffer(LightTexture lightTexture, float tickDelta, double d, double e, double g, CallbackInfo callback) {
```
coderbot16, line 191: This injection functions properly, however it might be nice to make sure that it only applies to the first depthMask call in this function. While renderWorldBounds sets up its expected depthMask state anyways, it's nice to be as precise as possible with our injections. For targeting the first injection, an ordinal=0 is acceptable.

coderbot16 (changes requested):
One minor comment about the injection but this is looking good otherwise.

src/main/java/net/coderbot/iris/mixin/MixinLevelRenderer.java
IMS212: Can we not modify variable?

coderbot16: You might be thinking of ModifyArg. I suppose that's possible too.

maximumpower55:
Made requested changes

coderbot16 (approved):
Thanks!

PR #1105, opened 2021-12-17
change mention that Sodium isn't bundled in FAQ

Andre601:
Sodium isn't bundled any longer in Iris, but the FAQ still mentioned that, so I updated this, while also linking to the Sodium modrinth page


coderbot16:
Thanks for the PR! I chose to word this differently but nice catch.

Andre601:
> Thanks for the PR! I chose to word this a tiny bit differently but nice catch.

You could've suggested a change, but okay. As long as the FAQ got updated is it fine by me.

Issue #300, opened 2021-05-27
Duplicated uniforms are not caught or typechecked properly


coderbot16:
This is because we insert into the locations / uniformNames map without checking whether they already have something with the given key

elwinmartin:
This looks like this is still an issue, is it worth doing this?

I have made some changes to check the hashmap on a branch and was wondering if it was worth opening a PR for this.

coderbot16:
Reopening until we release 1.1.4 with this fix.

PR #1111, opened 2021-12-19
Stop duplicated uniforms from being added

maximumpower55:
Will close: #300

### DIFF
--- a/src/main/java/net/coderbot/iris/gl/program/ProgramUniforms.java
+++ b/src/main/java/net/coderbot/iris/gl/program/ProgramUniforms.java
@@ -162,7 +162,7 @@ public class ProgramUniforms {
 
 			if (!locations.containsKey(id) && !uniformNames.containsKey(name)) {
 				locations.put(id, name);
-				externalUniformNames.put(name, type);
+				uniformNames.put(name, type);
 			} else {
 				Iris.logger.warn("[" + this.name + "] Duplicate uniform: " + type.toString().toLowerCase() + " " + name);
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/coderbot/iris/gl/program/ProgramUniforms.java
```diff
@@ -155,8 +160,14 @@ public OptionalInt location(String name, UniformType type) {
 				return OptionalInt.empty();
 			}
 
-			locations.put(id, name);
-			uniformNames.put(name, type);
+			if (!locations.containsKey(id) && !uniformNames.containsKey(name)) {
```
coderbot16, line 163: Did you mean uniformNames?

coderbot16, line 163: The only purpose of externally managed uniforms is just to suppress warnings for uniforms not managed through the uniform system - there isn't a pressing need to check them, but if we do, we should throw a hard error instead. Not sure if we should check if a normal uniform collides with an externally managed uniform.

coderbot16 (changes requested):

coderbot16:
Huh, looks like IntelliJ's built in PR review thing messed up the comment locations. In any case the first comment targeted the first check & the second targeted the second check.

src/main/java/net/coderbot/iris/gl/program/ProgramUniforms.java
maximumpower55: Yes

maximumpower55:
Made requested changes

coderbot16 (approved):
Thanks!

PR #1130, opened 2021-12-24
Updated comments about minor issues with shaders

Banzobotic:
Updated some comments about minor issues with some shaders in the supported shader list. 

Tested on 1.18 with commit eb7aabe.


coderbot16 (approved):
Thanks!

PR #1139, opened 2021-12-28
Add support for the prepare stage

maximumpower55:


src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
```diff
@@ -199,6 +200,11 @@ public DeferredWorldRenderingPipeline(ProgramSet programs) {
 			return shadowMapRenderer;
 		};
 
+		this.prepareRenderer = new CompositeRenderer(programs.getPackDirectives(), programs.getPrepare(), renderTargets,
```
coderbot16, line 203: `gbuffers_terrain` might be reading from the wrong flipped buffers with this current behavior. We'll need to split `flippedBeforeTranslucent` into `flippedBeforeShadow` and `flippedAfterPrepare`, and then update references accordingly. Also, `flippedBeforeTerrain` references need to be updated to either `flippedBeforeShadow` or `flippedAfterPrepare` as appropriate.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
maximumpower55: Will try to implement this today

maximumpower55: Fixed.

coderbot16:
I think you might have swapped the names of flippedBeforeShadow and flippedAfterPrepare.

maximumpower55:
oops i did

maximumpower55:
wait where is the area i swapped them or did you mean that i swapped the names of the variables

coderbot16:
The variable names

maximumpower55:
ah

maximumpower55:
Did i fix the swap correctly

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
```diff
@@ -194,13 +194,18 @@ public DeferredWorldRenderingPipeline(ProgramSet programs) {
 
 		this.centerDepthSampler = new CenterDepthSampler(renderTargets, updateNotifier);
 
-		flippedBeforeTranslucent = flipper.snapshot();
+		flippedAfterPrepare = flipper.snapshot();
```
coderbot16, line 197: You're taking the flippedAfterPrepare snapshot before the prepare renderer has been initialized? That doesn't seem right.

coderbot16 (changes requested):
One comment, otherwise seems good

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
maximumpower55: Fixed.

maximumpower55:
Made requested changes

coderbot16 (approved):
Thanks!

PR #1141, opened 2021-12-29
Fix closure syntax error in build.gradle

douira:
This is a syntax error, at least my gradle reports the following issue:
```
Ambiguous expression could be either a parameterless closure expression or an isolated open code block;
   solution: Add an explicit closure parameter list, e.g. {it -> ...}, or force it to be treated as an open block by giving it a label, e.g. L:{...} @ line 199, column 32.
```


coderbot16 (approved):
Thanks!

PR #1142, opened 2021-12-30
Use glsl-transformer for shader patching

douira:
This PR adds an abstract system for shader patching through `Patcher` and re-implements `TriforcePatcher` in `TransformPatcher` (as Triforce 2) using `glsl-transformer` for program transformation of the shaders.


IMS212:
"your actions shall have unforeseen consequences"

coderbot16:
I was just trying to rename the branch lol

PR #1143, opened 2021-12-30
The Big OpenGL Wrapper Refactor

maximumpower55:
This will allow for modern OpenGL features (Compute shaders, etc) to be implemented easier

Currently doesn't compile or run

Currently working on part 2 of this refactor which will make it work

After it compiles and runs i will add a list of tasks to be completed before this is ready for review 

Currently this is just a port of sodium's thingl updates to iris


maximumpower55:
Closing this for now

PR #1144, opened 2021-12-30
Fix Korean translation isn't working

craftingmod:
The correct Korean language code is `ko-kr` not `kr-kr`.

That's why no one could see Korean translation in game.

This commit fixes that.

> `ko-kr` is South Korea's Korean code and `ko-kp` is North Korea's Korean code(well no one would play minecraft in North Korea)


IMS212:
Thanks!

Issue #860, opened 2021-09-10
If graphics are set to Fabulous when shaders are disabled, enabling and then disabling shaders resets graphics option to Fancy


CalXee:
### What happened?

Title says it all. It's just annoying for the minority who play on Fabulous. https://youtu.be/I1IOr_w8iko

### Iris and Minecraft Version

Minecraft 1.17.1, Iris 1.1.2

### Are you running Sodium along with Iris?

Yes

### Operating System

Windows 10

### What is your GPU?

GTX 1050 Ti

### Java Version

Java 16

jmaximusix:
As far as I know the "Fabulous" option isn't compatible with shaders in general (same for optifine). So it might be intended behaviour to reset this option

Mr-Waterscout:
Yeah, Optifine gives you a message saying you can't run shaders with the "Fabulous" graphics option. I'm surprised Iris didn't tell you that along with changing the graphics option to "Fancy".

AlsoNiko:
This is definitely not an intended behavior. No, the shaders aren't running on fabulous, it auto sets to fancy and disables fabulous when shaders are enabled, however enabling shaders SHOULD revert back to fabulous, not fancy.

IMS212:
The issue here is enabling shaders resets your previous option to Fancy from Fabulous, instead of reverting to what you had last.

peow:
im using Iris 1.7.0 on Fabric 1.20.1 and this problem still persists :(

PR #1159, opened 2022-01-02
Restore Fabulous graphics if they were enabled before shaders were on

NoComment1105:
Prior to this, if fabulous graphics were enabled and a shader was then enabled, fabulous graphics would become fancy as intended. But then when shaders were turned of fabulous was not restored. This PR fixes this. (#860)


IMS212:
#1160 :mald:

NoComment1105:
As commented else where this is a little over complicated and I will be reducing its complexity soon

PR #1160, opened 2022-01-02
Override the checks for Fabulous instead of disabling it

IMS212:
This allows users to disable and reenable shaders without their Fabulous choice getting ruined.

### DIFF
--- a/src/main/java/net/coderbot/iris/fantastic/FabulousInterface.java
+++ b/src/main/java/net/coderbot/iris/fantastic/FabulousInterface.java
@@ -1,0 +1,5 @@
+package net.coderbot.iris.fantastic;
+
+public interface FabulousInterface {
+	boolean wasFabulous();
+}
--- a/src/main/java/net/coderbot/iris/mixin/fabulous/MixinDisableFabulousGraphics.java
+++ b/src/main/java/net/coderbot/iris/mixin/fabulous/MixinDisableFabulousGraphics.java
@@ -2,9 +2,11 @@ package net.coderbot.iris.mixin.fabulous;
 
 import net.coderbot.iris.Iris;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Overwrite;
 import org.spongepowered.asm.mixin.Unique;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.Redirect;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import net.fabricmc.api.EnvType;
 import net.fabricmc.api.Environment;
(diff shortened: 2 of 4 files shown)


coderbot16:
This seems like it could be error prone with mods that check the graphics mode directly. Realistically, we have to make it so that the graphicsMode field is FANCY or FAST to be the most robust.

IMS212:
I don't think almost any mods check directly for this, mostly due to it being harder to access.

ImperatorStorm:
what was wrong with https://github.com/IrisShaders/Iris/pull/1159 ?

IMS212:
It was needlessly complicated and tried swapping modes in and out, while this just tricks the game into thinking it’s on fancy instead of fabulous when shaders are on.

NoComment1105:
Why Trick the game with something that could die when there already was a system that worked perfectly (totally not pissed off to the moon)

coderbot16:
- https://github.com/Bonono63/Create---Fabric-Edition/blob/94d03047a9fba5d4346e0be91353a8d7dbd3a709/src/main/java/com/mrbonono63/create/CreateClient.java#L232

- https://github.com/CumulusMC/Art-of-Alchemy/blob/74dd7a87e2b8f6d6a5ed477074fe16dc4fec41a7/src/main/java/com/cumulusmc/artofalchemy/render/RendererTank.java#L41

 - https://github.com/qouteall/ImmersivePortalsMod/blob/02f07c119bde54455e3a1bd00ad77ba4bc875601/imm_ptl_core/src/main/java/qouteall/imm_ptl/core/IPModMainClient.java#L44

Clearly, some mods are checking the graphics mode directly. And, even if that doesn't break here, the lack of robustness here isn't acceptable. This might break some hidden assumptions, and even if that only happens for one mod, it will manifest as an extremely hard to reproduce bug that will only happen in this fabulous-but-not-fabulous state.

IMS212:
Hm, I'll look into a solution for this.

IMS212:
Fixed

src/main/java/net/coderbot/iris/fantastic/FabulousInterface.java
```diff
@@ -0,0 +1,5 @@
+package net.coderbot.iris.fantastic;
```
coderbot16, line 1: Wrong package?

src/main/java/net/coderbot/iris/mixin/fabulous/MixinDisableFabulousGraphics.java
```diff
@@ -34,12 +37,22 @@
...
+				options.graphicsMode = GraphicsStatus.FABULOUS;
+				this.wasFabulous = false;
+			}
 			return;
 		}
 
 		if (options.graphicsMode == GraphicsStatus.FABULOUS) {
 			// Disable fabulous graphics when shaders are enabled.
 			options.graphicsMode = GraphicsStatus.FANCY;
+			this.wasFabulous = true;
```
coderbot16, line 50: Won't this cause wasFabulous to get set to `true` if the user is cycling from Fancy to Fast  in video settings?

src/main/java/net/coderbot/iris/mixin/fabulous/MixinEnableFabulousGraphics.java
```diff
@@ -0,0 +1,34 @@
...
+	public GraphicsStatus graphicsMode;
+
+	@Inject(method = "save", at = @At("HEAD"))
+	private void onSave(CallbackInfo ci) {
+		if (((FabulousInterface) this.minecraft.levelRenderer).wasFabulous()) {
+			this.graphicsMode = GraphicsStatus.FABULOUS;
+		}
+	}
+
+	@Inject(method = "save", at = @At("TAIL"))
```
coderbot16, line 28: `TAIL` generally isn't needed unless there's a specific reason to not target all return statements.

```suggestion
	@Inject(method = "save", at = @At("RETURN"))
```

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/fantastic/FabulousInterface.java
CalXee: Tbh I thought the "fantastic" was an intentional funny name... but yea it's fabulous right?

PR #1161, opened 2022-01-03
blendFunc/vec4i uniform support

IMS212:


src/main/java/net/coderbot/iris/gl/blending/BlendModeStorage.java
```diff
@@ -32,6 +34,10 @@ public static void overrideBlend(BlendMode override) {
 		}
 
 		blendLocked = true;
+
+		if (blendFuncListener != null) {
+			blendFuncListener.run();
```
coderbot16, line 39: This isn't correct. This needs to be run at the end of `GlStateManager`'s blendFunc or blendFuncSeparate functions, not here - the blendFunc uniform isn't just for blend mode overrides.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/gl/blending/BlendModeStorage.java
IMS212: Fixed.

coderbot16:
Still not correct - HEAD of these functions is before vanilla sets the new blend mode, and also this should be a separate mixin instead of being stuffed into MixinGlStateManager_BlendModeOverride. Should inject at RETURN instead.

IMS212:
Should I move it to the State listening mixin?

coderbot16:
Sure

IMS212:
Done

src/main/java/net/coderbot/iris/mixin/MixinGlStateManager_BlendOverride.java
```diff
@@ -2,13 +2,15 @@
 
 import com.mojang.blaze3d.platform.GlStateManager;
 import net.coderbot.iris.gl.blending.BlendModeStorage;
+import net.coderbot.iris.gl.state.StateUpdateNotifiers;
```
coderbot16, line 5: Can roll back the unnecessary diff on this file

src/main/java/net/coderbot/iris/gl/blending/BlendModeStorage.java
```diff
@@ -1,6 +1,7 @@
 package net.coderbot.iris.gl.blending;
 
 import com.mojang.blaze3d.platform.GlStateManager;
+import net.coderbot.iris.gl.state.StateUpdateNotifiers;
```
coderbot16, line 4: Same here

coderbot16 (changes requested):

IMS212:
Fixed

coderbot16 (approved):
Thanks!

PR #1163, opened 2022-01-03
Fix shadow distance resetting after each boot, #852

### LINKED ISSUE #852: Max Shadow Distance resetting on each boot
### What happened?

I use sildurs EVL 1.29 Shaders, and I noticed that the max-shadow distance is always set to 32.
It isn't greyed out, so I always drop it down to 8/9, but each boot, this value resets to 32.
This is honestly very annoying, lol

### Screenshots

image

### Relevant log output

_No response_

iris-mc1.17-1.1.2.jar

Yes

### Operating System

Windows 10

### What is your GPU?

GTX 1070

### Java Version

16

### Additional context

_No response_

NoComment1105:
This Pull Request fixes #852, where the shadow distance slider, if changed in the sodium video settings menu only, would not be written to the config file, causing it to reset to the default of `32` after a restart


maximumpower55 (approved):
LGTM

coderbot16 (approved):
Thanks!

PR #1165, opened 2022-01-04
Add a workaround for when OptiFine has set the maxFps to zero in options.txt

maximumpower55:


src/main/java/net/coderbot/iris/mixin/MixinMaxFpsCrashFix.java
```diff
@@ -0,0 +1,32 @@
...
+import net.minecraft.client.Options;
+
+/**
+ * A workaround for when OptiFine has set the maxFps to zero in options.txt
+ * 
+ * Fun.
+ */
+@Mixin(Options.class)
+public abstract class MixinMaxFpsCrashFix {
+	@Redirect(
```
coderbot16, line 17: Might want to add an `allow = 1` to ensure that this only matches a single injection.

maximumpower55: Added.

coderbot16 (approved):
Thanks!

PR #1170, opened 2022-01-06
entityColor -> overlayColor attribute transformation on 1.16

IMS212:
This lets us get rid of all the bad entityColor hit flash code in favor of Vanilla's systems.
This does not support geometry shaders right now, though I don't know any geometry shaders that use entityColor.


src/main/java/net/coderbot/iris/pipeline/AttributeShaderTransformer.java
```diff
@@ -0,0 +1,41 @@
...
+			return null;
+		}
+
+		StringTransformations transformations = new StringTransformations(source);
+
+		//Add entity color -> overlay color attribute support.
+		// TODO: We don't handle the geometry shader here.
+		if (type == ShaderType.VERTEX) {
+			transformations.injectLine(Transformations.InjectionPoint.BEFORE_CODE, "uniform sampler2D overlay;");
+			transformations.replaceExact("uniform vec4 entityColor;", "");
```
coderbot16, line 19: `replaceExact` is too fragile here, because any whitespace will break it.

src/main/java/net/coderbot/iris/pipeline/AttributeShaderTransformer.java
```diff
@@ -0,0 +1,41 @@
...
+			transformations.replaceExact("main", "irisMain");
+			transformations.injectLine(Transformations.InjectionPoint.END, "void main() {\n" +
+					"	vec4 overlayColor = texture2D(overlay, (gl_TextureMatrix[2] * gl_MultiTexCoord2).xy);\n" +
+					"	overlayColor.a = 1.0 - overlayColor.a;\n" +
+					"	entityColor = overlayColor;\n" +
+					"\n" +
+					"    irisMain();\n" +
+					"}");
+		} else {
+			transformations.replaceExact("uniform vec4 entityColor;", "varying vec4 entityColor;");
```
coderbot16, line 36: replaceExact is too fragile here.

src/main/java/net/coderbot/iris/pipeline/AttributeShaderTransformer.java
```diff
@@ -0,0 +1,41 @@
...
+				throw new IllegalStateException("Shader already contains \"irisMain\"???");
+			}
+
+			// Create our own main function to wrap the existing main function, so that we can pass through the overlay color at the
+			// end.
+			transformations.replaceExact("main", "irisMain");
+			transformations.injectLine(Transformations.InjectionPoint.END, "void main() {\n" +
+					"	vec4 overlayColor = texture2D(overlay, (gl_TextureMatrix[2] * gl_MultiTexCoord2).xy);\n" +
+					"	overlayColor.a = 1.0 - overlayColor.a;\n" +
+					"	entityColor = overlayColor;\n" +
```
coderbot16, line 31:
```suggestion
					"	entityColor = vec4(overlayColor.rgb, 1.0 - overlayColor.a);\n" +
```

src/main/java/net/coderbot/iris/pipeline/AttributeShaderTransformer.java
```diff
@@ -0,0 +1,41 @@
...
+import net.coderbot.iris.shaderpack.transform.Transformations;
+
+public class AttributeShaderTransformer {
+	public static String patch(String source, ShaderType type) {
+		if (source == null) {
+			return null;
+		}
+
+		StringTransformations transformations = new StringTransformations(source);
+
```
coderbot16, line 14: should check for `iris_` prefix here, according to the other change

src/main/java/net/coderbot/iris/pipeline/AttributeShaderTransformer.java
```diff
@@ -0,0 +1,41 @@
+package net.coderbot.iris.pipeline;
+
+import net.coderbot.iris.gl.shader.ShaderType;
+import net.coderbot.iris.shaderpack.transform.StringTransformations;
+import net.coderbot.iris.shaderpack.transform.Transformations;
+
+public class AttributeShaderTransformer {
+	public static String patch(String source, ShaderType type) {
+		if (source == null) {
+			return null;
```
coderbot16, line 10: Null sources should not be passed to this function to begin with

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
```diff
@@ -513,10 +514,19 @@ private Pass createPass(ProgramSource source) {
 		Objects.requireNonNull(source.getVertexSource());
 		Objects.requireNonNull(source.getFragmentSource());
 		ProgramBuilder builder;
+		String vertex = source.getVertexSource().orElse(null);
+		String geometry = source.getGeometrySource().orElse(null);
+		String fragment = source.getFragmentSource().orElse(null);
+
+		if (source.getName().contains("entities")) {
```
coderbot16, line 521: This is incorrect, entityColor is available to all programs, but it will only have a value not equal to `vec4(0.0)` in entity rendering - which might not use `gbuffers_entities`. This code will need to be able to keep track of whether the overlay vertex attribute is in use, and only use the patched program variant if it is.

src/main/java/net/coderbot/iris/samplers/IrisSamplers.java
```diff
@@ -115,6 +115,7 @@ public static boolean addShadowSamplers(SamplerHolder samplers, ShadowMapRendere
 	public static void addLevelSamplers(SamplerHolder samplers, AbstractTexture normals, AbstractTexture specular) {
 		samplers.addExternalSampler(TextureUnit.TERRAIN.getSamplerId(), "tex", "texture", "gtexture");
 		samplers.addExternalSampler(TextureUnit.LIGHTMAP.getSamplerId(), "lightmap");
+		samplers.addExternalSampler(TextureUnit.OVERLAY.getSamplerId(), "overlay");
```
coderbot16, line 118: This should probably be `iris_overlay` or something, because the `overlay` sampler isn't part of the shaders standard, and is only available on MC 1.15+. In addition, we should reject shaders in `AttributeShaderTransformer` if they contain `iris_` before patching.

coderbot16 (changes requested):
This is a good start, though there are quite a few things that would need to be resolved before this could be merged.

IMS212:
Everything should be fixed in the upcoming commit except the overlay tracking, how would I accomplish that?

IMS212:
Could we just check for entities in GBufferPrograms?

coderbot16:
Potentially - main thing to check is what programs can potentially be used when rendering entities / similar that should also have a hurt color applied - all I can think of currently would be entities, hand, and maybe (???) stuff like spider eyes or glint? A closer investigation is probably appropriate

IMS212:
I've thought of something - will try.

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
IMS212: is this better?

src/main/java/net/coderbot/iris/pipeline/AttributeShaderTransformer.java
```diff
@@ -0,0 +1,40 @@
...
+	public static String patch(String source, ShaderType type) {
+		if (source.contains("iris_")) {
+			throw new IllegalStateException("Shader is attempting to exploit internal Iris code!");
+		}
+
+		StringTransformations transformations = new StringTransformations(source);
+
+		//Add entity color -> overlay color attribute support.
+		// TODO: We don't handle the geometry shader here.
+		if (type == ShaderType.VERTEX) {
```
coderbot16, line 17: We should try to handle geometry shaders. If you can't figure out a way I have an idea.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
```diff
@@ -290,13 +300,16 @@ public DeferredWorldRenderingPipeline(ProgramSet programs) {
...
-		this.hand = programs.getGbuffersHand().map(this::createPass).orElse(texturedLit);
-		this.handTranslucent = programs.getGbuffersHandWater().map(this::createPass).orElse(hand);
-		this.glowingEntities = programs.getGbuffersEntitiesGlowing().map(this::createPass).orElse(entities);
-		this.glint = programs.getGbuffersGlint().map(this::createPass).orElse(textured);
-		this.eyes = programs.getGbuffersEntityEyes().map(this::createPass).orElse(textured);
+		this.hand = programs.getGbuffersHand().map(this::createEntityPass).orElse(texturedLit);
+		this.handNoOverlay = programs.getGbuffersHand().map(this::createPass).orElse(texturedLit);
+		this.handTranslucent = programs.getGbuffersHandWater().map(this::createEntityPass).orElse(hand);
+		this.glowingEntities = programs.getGbuffersEntitiesGlowing().map(this::createEntityPass).orElse(entities);
+		this.glint = programs.getGbuffersGlint().map(this::createEntityPass).orElse(textured);
```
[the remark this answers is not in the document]
IMS212: It is not.

src/main/java/net/coderbot/iris/pipeline/AttributeShaderTransformer.java
IMS212: Feel free, I don't know what to put as the array number.

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
```diff
@@ -279,9 +289,9 @@ public DeferredWorldRenderingPipeline(ProgramSet programs) {
 			return builder.build();
 		};
 
-		this.basic = programs.getGbuffersBasic().map(this::createPass).orElse(null);
-		this.textured = programs.getGbuffersTextured().map(this::createPass).orElse(basic);
-		this.texturedLit = programs.getGbuffersTexturedLit().map(this::createPass).orElse(textured);
+		this.basic = programs.getGbuffersBasic().map(this::createEntityPass).orElse(null);
```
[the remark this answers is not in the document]
IMS212: Will fix.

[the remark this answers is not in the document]
IMS212: Fixed.

src/main/java/net/coderbot/iris/pipeline/AttributeShaderTransformer.java
coderbot16: Since it's the same for all vertices, you can just use the first element (`[0]`), assuming you're referring to what I think you're referring to.

IMS212: Ah, then I think I can implement this.

IMS212: How can we check for the existence of the use of entityColor without catching false positives like "entityColorOutput"?

coderbot16: I don't think we need to check for its existence - we should just always pass it through in the geometry shader, and always make it available in the vertex shader, then the driver can optimize it out if needed.

IMS212: No, the issue I'm talking about is catching things with the name "entityColor" that don't actually refer to entityColor.

coderbot16: Why would we need to do that?

IMS212: Does what I've put look good?

IMS212:
This should be good to go now, minus some hacky geometry shader support that i've added

coderbot16:
This conflicts with trunk currently.

IMS212:
Fixed

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -7,7 +7,9 @@
 import net.coderbot.iris.JomlConversions;
 import net.coderbot.iris.gl.state.StateUpdateNotifiers;
 import net.coderbot.iris.gl.uniform.DynamicUniformHolder;
-import net.coderbot.iris.gl.uniform.UniformHolder;
+import net.coderbot.iris.gl.uniform.UniformHolder;=======
```
oliviathevampire, line 10: This seems to have to be fixed

IMS212: Fixed

coderbot16:
This won't be merged before 1.2.0 releases, but might show up in 1.2.1.

IMS212:
Can this be looked into for 1.2.1?

coderbot16:
There's still a few potential holes in the logic here, but I'm merging this anyways since it's such a nice improvement to the internals. Thanks!

PR #1175, opened 2022-01-07
Make all standard macros available in preprocessed properties files

maximumpower55:


maximumpower55:
Marking this as draft until i can fix the tests failling

douira:
I'd like to note that your GPG key seems to have expired.

maximumpower55:
I made this pr before i had a gpg key

PR #1178, opened 2022-01-07
Add blurry start/velocity Complementary uniforms as well as halfLifeDown

IMS212:


src/main/java/net/coderbot/iris/uniforms/transforms/SmoothedFloat.java
```diff
@@ -38,25 +38,23 @@
...
 		// For example, a half life of value of 2.0 is 4 ticks or 0.2 seconds
-		halfLife *= 0.1f;
-
-		// Compute the decay constant from the half life
-		// https://en.wikipedia.org/wiki/Exponential_decay#Measuring_rates_of_decay
-		// https://en.wikipedia.org/wiki/Exponential_smoothing#Time_constant
-		double timeConstant = halfLife / LN_OF_2;
-		// k = 1 / τ
-		this.decayConstant = (float) (1.0f / timeConstant);
+		this.halfLifeUp = (halfLifeUp * 0.1F);
```
coderbot16, line 56: Why not compute and store two decay constants? Could move the decay constant computation into a function to avoid duplication

coderbot16 (changes requested):
One comment, otherwise seems good

src/main/java/net/coderbot/iris/uniforms/transforms/SmoothedFloat.java
IMS212: Will do

IMS212: Done

coderbot16:
That's not what I requested? I said we should compute both decay constants in the constructor and pick between them in the update call

IMS212:
Ah, nevermind. One second

IMS212:
Done

coderbot16:
Looks good, thanks!

PR #1179, opened 2022-01-07
Update German translation

Levelowel:
Updates the German translation with support for the Shader Options GUI


douira (approved):
Looks great! see #1190 for comments on my version of this translation. This translation is better.

IMS212:
Thanks!

PR #1190, opened 2022-01-10
Update de_de

douira:
I've updated the German translation of the strings. The string `options.iris.shadowDistance.sodium_tooltip` doesn't exist in `en_us` anymore. Should it be removed? Other translations also still include it though. If anyone here also speaks German, please check my work. Especially another opinion on the following issue would be appreciated:

The translations `Shaderpaket-Einstellungen` and `Shaderpaket-Auswahl` don't follow normal German grammar rules which would translate them as `Shaderpaketeinstellungen` and `Shaderpaketauswahl` instead. However, the `-` construction was already used for `Shaderpaket-Ordner` so I've continued the pattern.

### DIFF
--- a/src/main/resources/assets/iris/lang/de_de.json
+++ b/src/main/resources/assets/iris/lang/de_de.json
@@ -15,6 +15,8 @@
   "options.iris.apply": "Anwenden",
   "options.iris.refresh": "Aktualisieren",
   "options.iris.openShaderPackFolder": "Shaderpaket-Ordner öffnen ...",
+  "options.iris.shaderPackSettings": "Shaderpaket-Einstellungen...",
+  "options.iris.shaderPackList": "Shaderpaket-Auswahl...",
   "options.iris.refreshShaderPacks": "Shaderpakete aktualisieren",
   "options.iris.shaderPackSelection": "Shaderpakete ...",
   "options.iris.shaderPackSelection.title": "Shaderpakete",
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -15,6 +15,8 @@
   "options.iris.apply": "Anwenden",
   "options.iris.refresh": "Aktualisieren",
   "options.iris.openShaderPackFolder": "Shaderpaket-Ordner öffnen ...",
+  "options.iris.shaderPackSettings": "Shaderpaket-Einstellungen...",
```
Levelowel, line 18: Please include a space before the `...` at the end. I don't know why we did it that way but Optifine does the same as the previous translation so we should stay consistent

src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -15,6 +15,8 @@
...
+  "options.iris.shaderPackList": "Shaderpaket-Auswahl...",
```
Levelowel, line 19: Please also include a space here before the `...` at the end. 

I like the idea of an `Auswahl` but it's still not the same as list and the original is a list

src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -26,11 +28,22 @@
   "options.iris.shaderPackSelection.copyErrorAlreadyExists": "\"%s\" befindet sich bereits im Shaderpaket-Ordner!",
   "options.iris.shaders.disabled": "Shader: Deaktiviert",
   "options.iris.shaders.enabled": "Shader: Aktiviert",
+  "options.iris.back": "Zurück",
+  "options.iris.reset": "Zurücksetzen",
+  "options.iris.reset.tooltip": "Setze ALLE Einstellungen zurück und wende diese an?",
```
Levelowel, line 33: This sounds a bit weird because you don't really have a subject in this question

src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -26,11 +28,22 @@
...
+  "options.iris.reset.tooltip.holdShift": "Halte SHIFT und klicke um zurückzusetzen",
```
Levelowel, line 34: It's actually called UMSCHALTTASTE in German.

`Halte` also doesn't quite fit. `halte ... gedrückt` seems better in this case.

This also sounds a bit weird because you don't really have a subject in this question as well

src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -26,11 +28,22 @@
...
+  "options.iris.setToDefault": "Setze Einstellung zurück auf Standardwert?",
```
Levelowel, line 35: Same goes for this sentence. It just sounds a bit weird when you don't have a proper subject

src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -26,11 +28,22 @@
...
+  "options.iris.profile": "Profil",
+  "options.iris.profile.custom": "Benutzerdefiniert",
   "options.iris.shadowDistance": "Max. Schattendistanz",
   "options.iris.shadowDistance.enabled": "Erlaubt es dir, die maximale Distanz für Schatten zu ändern. Gelände und Objekte, die weiter entfernt sind als die eingestellte Distanz, werden keine Schatten werfen. Das Senken der Distanz kann die Leistung erheblich verbessern.",
   "options.iris.shadowDistance.disabled": "Dein aktuelles Shaderpaket hat bereits eine Distanz für Schatten gesetzt; du kannst sie nicht ändern.",
   "options.iris.shadowDistance.sodium_tooltip": "Die Distanz für Schatten bestimmt, wie weit das Gelände entfernt sein darf, um keinen Schatten mehr zu werfen. Bei geringeren Distanzen werden weniger Schatten berechnet, was die Bildrate verbessert. Diese Option kann nicht geändert werden bei Paketen, die explizit eine Distanz für Schatten vorgeben. Die tatsächliche Distanz für Schatten wird von der Einstellung \"Sichtweite\" begrenzt.",
 
   "pack.iris.select.title": "Auswählen",
-  "pack.iris.list.label": "+ Um Shaderpakete hinzuzufügen, ziehe sie in dieses Fenster hinein"
+  "pack.iris.configure.title": "Configure",
```
Levelowel, line 44: You somehow forgot to translate this?

Levelowel (changes requested):
I've already made a pull request to update the German translation (#1179). Please check it out as well. `options.iris.shadowDistance.sodium_tooltip` is only used in 1.17 and 1.18. (but translations are merged into trunk that's why it's still there)

And regarding the other issue: Optifine uses `Shaderpaket-Ordner` and my translation also follows this pattern as I think it's generally easier to read when adding a `-`. Though I'm unsure if we should follow Optifine anyway because the translations include `Shaderpaketeinstellungen` and sometimes even `Shader-Paket` so I also want other Germans to report back what they prefer

douira:
Well if you've already translated it all then this PR is pointless. I checked for if there was already a PR for this open but wasn't looking for "German" but rather "de_de" in the title. I'll close this and instead have look at your PR

src/main/resources/assets/iris/lang/de_de.json
douira: good point

src/main/resources/assets/iris/lang/de_de.json
douira: yes I forgot.

PR #1191, opened 2022-01-10
Updated it_it.json (13 January 2022)

devpelux:
Updated it_it.json (10 January 2022), fixed some strings that exceeded the size of the buttons.


IMS212:
Thanks!

PR #1193, opened 2022-01-10
Define render stages and their ordinals

IMS212:
Render stages still don't work, but anything depending on the macros should compile now.


IMS212:
(This fixes SEUS PTGI HRR 3 compiling, but there are still weird unrelated issues)

coderbot16:
Closing in favor of #1195.

PR #1195, opened 2022-01-11
Render stage support (actually)

IMS212:
Mostly clean code, and I've confirmed all sky related render stages work. (Have not confirmed stages relating to outlines)

### DIFF
--- a/src/main/java/net/coderbot/iris/pipeline/WorldRenderingPhase.java
+++ b/src/main/java/net/coderbot/iris/pipeline/WorldRenderingPhase.java
@@ -1,6 +1,5 @@
 package net.coderbot.iris.pipeline;
 
-// TODO: Implement this! Right now, this is just here as a stub so packs depending on the macros compile.
 public enum WorldRenderingPhase {
 	NONE,
 	SKY,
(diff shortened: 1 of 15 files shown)


IMS212:
Yes, this fixes SEUS PTGI HRR 3.

IMS212:
Everything works except for outline and destroy, will look into this.

src/main/java/net/coderbot/iris/pipeline/WorldRenderingPhase.java
```diff
@@ -0,0 +1,29 @@
+package net.coderbot.iris.pipeline;
+
+// TODO: Implement this! Right now, this is just here as a stub so packs depending on the macros compile.
```
coderbot16, line 3: This TODO seems outdated...

coderbot16 (changes requested):
Looks like this conflicts, and I have a few minor comments. Other than that this looks very nice and is a great improvement over the initial PR.

IMS212:
I've yet to find out the cause of outline and destroy, it seems like we need to set those as the program gets used.

coderbot16:
Oh right, anything using the BufferSource can't be done by just injecting around calls in LevelRenderer.

IMS212:
Fixed.

coderbot16 (approved):
Looks good, thanks!

Issue #893, opened 2021-09-25
Iris should support occlusion culling in the shadow pass somehow


coderbot16:
Advanced Shadow Frustum Culling in Iris 1.1.2 is pretty much the best we can do with normal frustum culling techniques. Occlusion culling is needed to improve culling. Obviously, it doesn't work with voxelization. Though, it does work with the sun-bounce GI in SEUS.

IMS212:
Some info about how Canvas does this was explained by grondag in the caffeine server:
`Canvas uses two additional occlusion rasters for shadow pass - the first (lets call it A) is visible world geometry in ortho projection. Occluders that are drawn to the perspective-projection rasterizer are also drawn to A.  The second raster (B) contains shadowing occluders, also drawn in ortho projection.  Occluders in B are sorted starting from the light source. (Ortho sorting can be less expensive than perspective - it is essentially step-wise iteration in three dimensions.)

Each potential shadow occluder is first compared against A.  (You could still use your nice extruded frustum test as a pre-filter, though I haven't tried that yet.) If the occluder doesn't intersect with any part of A, then it cannot occlude anything in view and is discarded.  If it does intersect with A, then we try to draw it on B.  If all covered pixels were already filled in B, then the occluder was already fully in shadow and can be discarded.  If any pixel was filled, then the occluder should be part of the shadow pass. 

This is efficient but not perfect.  If you are standing in a fully-enclosed cave looking towards the sun (which you cannot see), the process will still iterate from the sun all the way down to the exterior surface of the cave.  It does give the correct result - that your view is entirely in shadow - but it seems relatively expensive in the nominal case. `

PR #1206, opened 2022-01-15
Add perspective projection matrix support

IMS212:
Simple code.

I have not tested this on an actual pack yet, however the matrix completely matches OptiFine's as well as Mojang's "perspective" function.

### DIFF
--- a/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
+++ b/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
@@ -512,7 +512,7 @@ public class ShadowRenderer implements ShadowMapRenderer {
 		float[] projMatrix;
 		if (this.fov != null) {
 			// If FOV is not null, the pack wants a perspective based projection matrix. (This is to support legacy packs)
-			projMatrix = ShadowMatrices.createPerspectiveMatrix(this.fov, resolution / resolution);
+			projMatrix = ShadowMatrices.createPerspectiveMatrix(this.fov);
 		} else {
 			projMatrix = ShadowMatrices.createOrthoMatrix(halfPlaneLength);
 		}
--- a/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
+++ b/src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
@@ -2,7 +2,6 @@ package net.coderbot.iris.pipeline;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.mojang.blaze3d.platform.GlStateManager;
 import com.mojang.blaze3d.shaders.ProgramManager;
 import com.mojang.blaze3d.systems.RenderSystem;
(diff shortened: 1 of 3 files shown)


IMS212:
Thanks- wait a second

Though seriously, this is pretty simple, would someone be able to make a test pack?

coderbot16:
Honestly I'm not sure if any packs made in the past decade actually use FOV in the shadow pass. Seems likely that it was just legacy silliness.

src/main/java/net/coderbot/iris/shadow/ShadowMatrices.java
```diff
@@ -23,6 +23,21 @@
 		};
 	}
 
+	public static float[] createPerspectiveMatrix(float fov, float aspect) {
+		float i = (float)(1.0D / Math.tan(fov * 0.01745329238474369D / 2.0D));
+		Matrix4f matrix4f = new Matrix4f();
+		return new float[] {
```
coderbot16, line 29: This code gives me the feeling that it was copied from somewhere, with the arbitrary constants and the unused Matrix4f. I'd prefer if this code weren't copied & had more documentation, such as what units FOV has.

src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -75,8 +72,9 @@
 	private final float renderDistanceMultiplier;
 	private final int resolution;
 	private final float intervalSize;
+	private final Float fov;
```
coderbot16, line 75: What units is FOV in? (degrees or radians) - this is important to document in all the places we have an FOV field.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/shadow/ShadowMatrices.java
IMS212: Correct, it was copied from Vanilla's projection method. I'll look into this.

coderbot16:
https://github.com/IrisShaders/Iris/issues/893

IMS212:
Fixed.

src/main/java/net/coderbot/iris/shadow/ShadowMatrices.java
```diff
@@ -23,6 +23,21 @@
 		};
 	}
 
+	public static float[] createPerspectiveMatrix(float fov) {
+		// This converts from degrees to radians.
+		float rad = (float) (1.0f / Math.tan(Math.toRadians(fov) * 0.5f));
```
coderbot16, line 28: This concerts to radians, but just to call tan(). A more accurate name is yScale.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/shadow/ShadowMatrices.java
IMS212: Done

coderbot16 (approved):
Thanks!

PR #1209, opened 2022-01-17
Update zh_tw.json

HJ-zhtw:
* Some subtle adjustments


src/main/resources/assets/iris/lang/zh_tw.json
```diff
@@ -20,21 +20,21 @@
   "options.iris.refreshShaderPacks": "重新整理",
   "options.iris.shaderPackSelection": "光影包……",
   "options.iris.shaderPackSelection.title": "光影包",
-  "options.iris.shaderPackSelection.addedPack": "已新增光影包 \"%s\"",
+  "options.iris.shaderPackSelection.addedPack": "已新增「\"%s\"」光影包",
```
coderbot16, line 23: Isn't it a little redundant to have both types of quotes here?

src/main/resources/assets/iris/lang/zh_tw.json
```diff
@@ -44,11 +44,11 @@
   "options.iris.shadowDistance": "最大陰影距離",
   "options.iris.shadowDistance.enabled": "能讓您變更陰影的最大距離。超出此距離的地形和實體將不會有陰影。調低陰影距離能顯著提升效能。",
   "options.iris.shadowDistance.disabled": "您無法變更目前光影包所設定的陰影顯示距離。",
-  "options.iris.shadowDistance.sodium_tooltip": "「陰影顯示距離」選項能控制在陰影傳遞（shadow pass）時，可能呈現的地形有多遠。 較近的距離將呈現較少的地形，但畫面播放速率（frame rates）也得以增加。 此選項在套用「有特定陰影顯示距離的光影包」時無法變更。 實際的陰影顯示距離會受「顯示距離」選項限制。",
+  "options.iris.shadowDistance.sodium_tooltip": "「陰影顯示距離」選項能控制在陰影傳遞（ shadow pass）時，可能呈現的地形有多遠。 較近的距離將呈現較少的地形，但畫面播放速率（frame rates）也得以增加。 此選項在套用「有特定陰影顯示距離的光影包」時無法變更。 實際的陰影顯示距離會受「顯示距離」選項限 制。",
```
coderbot16, line 47: what's the need for the extra space in the parentheses?

src/main/resources/assets/iris/lang/zh_tw.json
HJ-zhtw: Frankly, maybe.
Since we don't use the type used in English, I think it will be more appropriate to add the Chinese type.
If it might cause confusion or something, `「` and `」` can be replaced by space.

src/main/resources/assets/iris/lang/zh_tw.json
HJ-zhtw: For readability.
To show the difference, screenshots attached
without
* Without the extra space.  The "w" of the word shadow goes to the newline. The Chinese period does that too.

src/main/resources/assets/iris/lang/zh_tw.json
coderbot16: Would it then be natural to remove the English type quotes and only leave the Chinese type quotes?

src/main/resources/assets/iris/lang/zh_tw.json
coderbot16: Thanks for the explanation, makes sense.

src/main/resources/assets/iris/lang/zh_tw.json
HJ-zhtw: Unfortunately, it doesn't apply the translations file if doing so.

HJ-zhtw: Nope.

HJ-zhtw: > Did you remove the backslashes too when you removed the quotes? Otherwise the escaping breaks.

I didn't remove them. I left the backslashes.

douira: You'd have to remove the backslashes. If there are backslashes but no quotes then the file is invalid.

HJ-zhtw: > You'd have to remove the backslashes. If there are backslashes but no quotes then the file is invalid.

Just tried it with 1.2.0-rc1. The translations file didn't work properly.

coderbot16 (approved):
Thanks!

PR #1215, opened 2022-01-18
Update German translation

Levelowel:


src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -2,36 +2,42 @@
...
 
   "options.iris.apply": "Anwenden",
   "options.iris.refresh": "Aktualisieren",
-  "options.iris.openShaderPackFolder": "Shaderpaket-Ordner öffnen ...",
-  "options.iris.shaderPackSettings": "Shaderpaket-Einstellungen ...",
-  "options.iris.shaderPackList": "Shaderpaket-Liste ...",
-  "options.iris.refreshShaderPacks": "Shaderpakete aktualisieren",
+  "options.iris.openShaderPackFolder": "Shader-Ordner öffnen ...",
+  "options.iris.shaderPackSettings": "Shader-Einstellungen ...",
+  "options.iris.shaderPackList": "Shader-Liste ...",
```
douira, line 19: Why is there a distinction between "shader" and "shaderpaket" here? Sometimes you write the one and sometimes you write the other. In the English original it's always "shaderpack" except for when "shaders" are referenced as a feature in general. (see the first few lines of the file) It's actually pretty consistent in the original I think.

src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -2,36 +2,42 @@
...
+  "options.iris.refreshShaderPacks": "Shader/Liste aktualisieren",
```
douira, line 20: In the English version this also doesn't say "Liste". What is your reasoning here?

src/main/resources/assets/iris/lang/de_de.json
```diff
@@ -2,36 +2,42 @@
...
+  "options.iris.shaderPackSelection.failedAdd": "Die Dateien waren keine zulässigen Shaderpakete",
+  "options.iris.shaderPackSelection.failedAddSingle": "\"%s\" ist kein zulässiges Shaderpaket",
+  "options.iris.shaderPackSelection.copyError": "Konnte Shaderpaket \"%s\" nicht hinzufügen",
+  "options.iris.shaderPackSelection.copyErrorAlreadyExists": "\"%s\" befindet sich bereits im Shader-Ordner!",
+  "options.iris.shaderPackOptions.tooManyFiles": "Es können nicht mehrere Shader-Einstellungsdateien importiert werden!",
+  "options.iris.shaderPackOptions.failedImport": "Shader-Einstellungsdatei \"%s\" konnte nicht importiert werden",
+  "options.iris.shaderPackOptions.importedSettings": "Shader-Einstellungen aus \"%s\" importiert",
   "options.iris.shaders.disabled": "Shader: Deaktiviert",
   "options.iris.shaders.enabled": "Shader: Aktiviert",
+  "options.iris.shaders.nonePresent": "Shader: Keine Shader vorhanden",
```
douira, line 34: I find this a little confusing

douira (changes requested):
Just wanted to give my opinion on some of the wording. (I'm a German native speaker.)

src/main/resources/assets/iris/lang/de_de.json
Levelowel: Because some of the strings don’t quite fit on the buttons and I‘ve adjusted them accordingly as well as those that follow a similar scheme

src/main/resources/assets/iris/lang/de_de.json
Levelowel: Pressing the button will also refresh the list to include shaderpacks that were manually added without drag and-drop and I wanted to imply that

src/main/resources/assets/iris/lang/de_de.json
Levelowel: Yeah, this one is a little weird but I don’t like the idea of a “Paket” in general so I will remove the second Shader here

src/main/resources/assets/iris/lang/de_de.json
douira: Just looking it at, it feels weird, but if this is required to fit the long words into the buttons it's good.

src/main/resources/assets/iris/lang/de_de.json
douira: That makes sense. I guess the English version kind implies that the list is being refreshed without saying it. This way is good too.

src/main/resources/assets/iris/lang/de_de.json
douira: Sounds good

src/main/resources/assets/iris/lang/de_de.json
Levelowel: Yeah I know but it’s necessary (IMS allowed me to do that) and optifine as well as vanilla do the same (I really dislike the German translation for vanilla minecraft because it includes so many weird sounding words and way to many abbreviations)

douira: yeah the German translation for Minecraft feels somewhat clunky and unnatural at times. But I think at the same time that means the standard for how natural German translations have to sound isn't very high.

src/main/resources/assets/iris/lang/de_de.json
Levelowel: I just want to make it a little easier to understand what it actually does because “aktualisieren” also means update and users may press the button to update their shaderpacks then. This way that will less likely be the case

douira (approved):

coderbot16:
Thanks!

PR #1231, opened 2022-01-21
Update to 1.18.2

IMS212:
Ready.


src/main/java/net/coderbot/iris/mixin/MixinLevelRenderer.java
```diff
@@ -123,33 +124,33 @@
...
 	@Inject(method = "renderSky", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/player/LocalPlayer;getEyePosition(F)Lnet/minecraft/world/phys/Vec3;"))
-	private void iris$setVoidRenderStage(PoseStack poseStack, Matrix4f matrix4f, float f, Runnable runnable, CallbackInfo ci) {
+	private void iris$setVoidRenderStage(PoseStack poseStack, Matrix4f matrix4f, float f, Camera camera, boolean bl, Runnable runnable, CallbackInfo ci) {
 		pipeline.setPhase(WorldRenderingPhase.VOID);
 	}
 
 	@Inject(method = "renderSky", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/multiplayer/ClientLevel;getTimeOfDay(F)F"),
 			slice = @Slice(from = @At(value = "FIELD", target = "com/mojang/math/Vector3f.YP : Lcom/mojang/math/Vector3f;")))
-	private void iris$renderSky$tiltSun(PoseStack poseStack, Matrix4f projectionMatrix, float f, Runnable runnable, CallbackInfo callback) {
+	private void iris$renderSky$tiltSun(PoseStack poseStack, Matrix4f matrix4f, float f, Camera camera, boolean bl, Runnable runnable, CallbackInfo ci) {
```
coderbot16, line 153: matrix4f -> projectionMatrix?

src/main/java/net/coderbot/iris/uniforms/FogUniforms.java
```diff
@@ -18,6 +19,7 @@ public static void addFogUniforms(DynamicUniformHolder uniforms, FogMode fogMode
 		if (fogMode == FogMode.OFF) {
 			uniforms.uniform1f(UniformUpdateFrequency.ONCE, "fogDensity", () -> 0.0F);
 			uniforms.uniform1i(UniformUpdateFrequency.ONCE, "fogMode", () -> 0);
+			uniforms.uniform1i(UniformUpdateFrequency.ONCE, "fogShape", FogShape.CYLINDER::getIndex);
```
coderbot16, line 22: I don't like the idea of not defining what this means ourselves & just passing through the vanilla enum, especially if the vanilla enum changes.

src/main/java/net/coderbot/iris/uniforms/FogUniforms.java
```diff
@@ -33,6 +35,8 @@ public static void addFogUniforms(DynamicUniformHolder uniforms, FogMode fogMode
 					return GL11.GL_EXP2;
 				}
 			}, notifier -> {});
+
+			uniforms.uniform1i(PER_FRAME, "fogShape", RenderSystem.getShaderFogShape()::getIndex);
```
coderbot16, line 39: Ditto.

coderbot16 (changes requested):

IMS212:
I've fixed the projectionMatrix name everywhere, as well as fogShape by defining 0 as cylindrical and 1 as spherical.

coderbot16:
I think it would make the most sense to define 0 to be spherical and 1 to be cylindrical. The uniform will be set to the default value 0 on older versions of Iris and OptiFine, which matches existing shader behavior - spherical fog.

IMS212:
Sounds good

coderbot16:
I imagine we can leave the default fogShape as 0 when fog is disabled, not that shaders should be relying on any value when fog is disabled.

IMS212:
Some shaders may depend on fogShape in the future when there’s no fog, so not sure. It might be safe to just keep it like this

coderbot16:
If there's no fog why would they need the shape of it?

IMS212:
Shader’s like Xordev’s default may end up assuming fogShape == 0 means “in water” (bad assumption, but still possible), so fogShape == 1 would counter that.

coderbot16:
Well fogShape means whatever we want it to mean so we don't have to deal with that kind of legacy weirdness

IMS212:
Can we default to -1 for no fog?

coderbot16:
That could work

IMS212:
This is scheduled to be merged on Feb. 27, one day before the release of 1.18.2.

coderbot16:
This doesn't look mergeable - it conflicts with the development branch & we don't know when the next Sodium version will be out.

IMS212:
Resolved the conflict, but Sodium is still an issue, yes.

coderbot16 (commented):
The changes here look fine but this cannot be merged until Sodium is updated. I guess people will just have to wait a bit to use shaders on 1.18.2, not a huge deal overall.

IMS212:
Should we still have this merged, or just keep the 1.18.2 branch as-is?

coderbot16:
We can keep the 1.18.2 branch and remove the 1.18 branch, since it's been superseded.

PR #1232, opened 2022-01-21
Add shaderpack name, profile and number of changed options to crash reports

altrisi:
This pull request adds a field to crash reports indicating the loaded shaderpack name, as well as the profile and modified options (as shown in F3).

imagen


coderbot16 (approved):
Thanks!

PR #1243, opened 2022-01-25
Doc

Lolothepro:


coderbot16:
It looks like this contains a bunch of unrelated changes to images and such. Could they be reverted? Also, if the bug you're removing is fixed in a given newer version of Sildur's Vibrant Shaders, you should update the version in the section name.

PR #1246, opened 2022-01-27
Update Turkish language

egeesin:
- Improve translation in the existing strings
- Translate shader option GUI strings


coderbot16 (approved):
Thanks!

PR #1257, opened 2022-01-31
Geometry shader support in vanilla 1.17!

IMS212:
Uses some hacks to add the geometry field before the super constructor, but is otherwise quite clean.


oliviathevampire:
Does this work? Like is geometry shaders used with this?

IMS212:
Yes.

src/main/java/net/coderbot/iris/pipeline/newshader/ExtendedShader.java
```diff
@@ -180,4 +185,31 @@ public void addTextureImage(IntSupplier textureID, InternalTextureFormat interna
 		// mark for rebuild if needed
 		this.currentImages = null;
 	}
+
+	@Override
+	public void attachToProgram() {
+		this.getFragmentProgram().attachToShader(this);
```
coderbot16, line 191: Why not just call super() instead of doing the fragment and vertex attaching ourselves?

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/pipeline/newshader/ExtendedShader.java
IMS212: I thought they had to be in order, but if not, sure.

IMS212:
Done

coderbot16:
Nice work, thanks!

PR #1259, opened 2022-01-31
Add support for the hand depth macro

maximumpower55:


coderbot16:
Should we make this a constant in a single place instead of hardcoding it in two places?

maximumpower55:
Sure would defining it in the main hand rendering class work?

coderbot16:
That sounds fine for now

maximumpower55:
> That sounds fine for now

Added.

coderbot16 (approved):
Thanks!

PR #1260, opened 2022-02-03
Extend entity vertex format (mostly) correctly

IMS212:
This has worked without issues on the PBR branch for a while now, so I figured we should move it to trunk.


src/main/java/net/coderbot/iris/mixin/vertices/MixinBufferBuilder.java
```diff
@@ -76,8 +76,8 @@
 	@Inject(method = "begin", at = @At("RETURN"))
 	private void iris$afterBegin(int drawMode, VertexFormat format, CallbackInfo ci) {
 		if (extending) {
-			this.format = IrisVertexFormats.TERRAIN;
-			this.currentElement = IrisVertexFormats.TERRAIN.getElements().get(0);
+			this.format = format.getElements().contains(DefaultVertexFormat.ELEMENT_UV1) ? IrisVertexFormats.ENTITY : IrisVertexFormats.TERRAIN;
```
coderbot16, line 79: Why not just check for reference equality with TERRAIN / ENTITY / BLOCK?

src/main/java/net/coderbot/iris/mixin/vertices/MixinBufferBuilder.java
```diff
@@ -76,8 +76,8 @@
 	@Inject(method = "begin", at = @At("RETURN"))
 	private void iris$afterBegin(int drawMode, VertexFormat format, CallbackInfo ci) {
 		if (extending) {
-			this.format = IrisVertexFormats.TERRAIN;
-			this.currentElement = IrisVertexFormats.TERRAIN.getElements().get(0);
+			this.format = format == DefaultVertexFormat.NEW_ENTITY ? IrisVertexFormats.ENTITY : IrisVertexFormats.TERRAIN;
```
coderbot16, line 79: If the format here is IrisVertexFormats.ENTITY this will give terrain, which isn't correct. You'll need more robust logic here

coderbot16 (approved):
Thanks! This should help with the weirdness with Advanced Materials on BSL as well.

PR #1261, opened 2022-02-04
Fix notes on setting up workspace

douira:
If my comment on vscode is wrong, I can remove it.


coderbot16 (approved):
Thanks!

BeansBeefBroccoli:
Why do you default to IDEA?

PR #1270, opened 2022-02-08
Fixes for Create (Fabric) / Flywheel

IMS212:
Current fixes:
Change level renderer buffers during shadow rendering
~~Iris API: getShadowFrustum (null when not rendering shadows)~~ removed for now


src/main/java/net/coderbot/iris/pipeline/ShadowRenderer.java
```diff
@@ -518,6 +518,9 @@ public void renderShadows(LevelRendererAccessor levelRenderer, Camera playerCame
 		profiler.popPush("shadows");
 		ACTIVE = true;
 
+		RenderBuffers playerBuffers = levelRenderer.getRenderBuffers();
```
coderbot16, line 521: Could we add a comment that this is needed for mods like Flywheel? Just good to note down why this code is needed so we don't delete it in a refactor.

IMS212: done

coderbot16:
This seems to conflict with the shadow rendering changes that I just merged.

IMS212:
Fixed

IMS212:
Do you think we can get this merged before 1.2.1?

coderbot16:
Yes I do believe we can get it merged before then :P

PR #1271, opened 2022-02-09
docs: Update development-builds.md

microstar2022:
Update the step of getting development builds.


coderbot16:
Thanks for pointing this out. I chose to make the change myself to fix up a small issue I noticed in the command syntax that I made when writing the original brachyura docs.

microstar2022:
OK, I got it.

PR #1272, opened 2022-02-09
docs: Update development-builds.md

microstar2022:
Update the step of getting development builds.


NoComment1105:
Thank you for the second pull request, however, we can easily bring your commit from trunk to the other branches required. Good spot on the original!

PR #1274, opened 2022-02-09
Add IrisLogger interface

IMS212:
This will allow us to easily move to SLF4J for 1.18.2 without breaking compatibility.


coderbot16 (approved):
Thanks!

PR #1288, opened 2022-02-22
New RenderTypeShardInterface to add temporary shards for entityId and blockEntityId

IMS212:
This allows for a temporary shard to be inserted into a render type, which gets cleared with the render type state. This is useful for entityId and blockEntityId.


IMS212:
This is one of the last uses of render type wrapping outside of programs.

coderbot16:
As discussed on discord - this doesn't actually work. It'll break if an entity and a block entity use the same render type in the same frame.

IMS212:
Yes, I'm working on an alternative in the experimental codebase.

PR #1293, opened 2022-02-26
fix buildscript for jdt

CoolMineman:
bottom text


IMS212:
cringe user

looks good to me

PR #1294, opened 2022-02-26
Remove un-used code in issue config 

NoComment1105:
Tiny PR. Small typo in the issue config I noticed while reading it.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,3 @@
-blank_issues_enabled: false
 contact_links:
   - name: Join the Discord for support using Iris!
     url: https://discord.gg/jQJnav2jPu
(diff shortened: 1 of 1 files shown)


IMS212:
Is this not intentionally enabled?

NoComment1105:
It was enabled before, however it does completely remove the new issue opening ability which coderbot likes, so I might remove the option entirely, depending on what coder says

coderbot16:
Being able to open blank issues is nice for developers to use the issue tracker as a more formal "TODO" list. I don't think users opening blank issues has been a problem. So we should probably just leave them enabled. Worst case, if a user opens a blank issue, we can always ask them to reopen using the template if they don't provide enough info.

NoComment1105:
Alright, I'm gonna reopen this with the line removed

NoComment1105:
Sorted

coderbot16 (approved):
Thanks!

PR #1303, opened 2022-03-05
Optional LWJGLX debugging toggled via iris.properties + client commands

IMS212:
PRing this so we can delete the branch and just have it. It’s just good.


src/main/java/net/coderbot/iris/config/IrisConfig.java
```diff
@@ -26,11 +26,17 @@
 	 */
 	private boolean enableShaders;
 
+	/**
+	 * If debug features should be enabled. Gives much more detailed GLSL error outputs at the cost of performance.
```
coderbot16, line 30:
```suggestion
	 * If debug features should be enabled. Gives much more detailed OpenGL error outputs at the cost of performance.
```

src/main/java/net/coderbot/iris/Iris.java
```diff
@@ -116,6 +117,11 @@ public static void onRenderSystemInit() {
 			return;
 		}
 
+		if (irisConfig.isDebugEnabled()) {
+			logger.warn("Debug features enabled, GLSL logging will be more verbose!");
```
coderbot16, line 121:
```suggestion
			logger.info("Debug features enabled, OpenGL error logging will be more verbose!");
```

src/main/java/net/coderbot/iris/gl/GLDebug.java
```diff
@@ -0,0 +1,300 @@
...
+		StackTraceElement[] elems = throwable.getStackTrace();
+		StackTraceElement[] filtered = new StackTraceElement[elems.length];
+		int j = 0;
+		for (int i = offset; i < elems.length; i++) {
+			String className = elems[i].getClassName();
+			if (className == null) {
+				className = "";
+			}
+			if (className.startsWith("org.lwjglx.") && !className.startsWith("org.lwjglx.debug.opengl")
+					&& !className.startsWith("org.lwjglx.debug.glfw")) {
```
coderbot16, line 51: This seems like something that needs to be updated.

coderbot16:
This doesn't appear to actually compile.

IMS212:
Fixed

IMS212:
New client commands have been added via /iris, notably to toggle debugging, if shaders are enabled, and reloading without hotkeys.

BeansBeefBroccoli:
Are the hotkeys going to stay?

coderbot16 (approved):
Looks good, thanks.

PR #1311, opened 2022-03-09
Use glsl-transformer for shader patching 1.18

douira:
This PR adds an abstract system for shader patching through `Patcher` and re-implements `TriforcePatcher` in `TransformPatcher` (as Triforce 2) using `glsl-transformer` for program transformation of the shaders.

Previous PR that was closed to rename the branch: https://github.com/IrisShaders/Iris/pull/1142


douira:
Closed in favor of #1312

PR #1335, opened 2022-03-24
Fixed translations to make the message more clear.

EmqsaGit:
Fixed some translations so the user can more easily understand what they mean. Yes, shaders may mean "Sombreadores", but in the topic of Minecraft Shaders, we use "Shaders" or "Shader Packs" instead of "Sombreadores" or "Pacotes de Sombreadores".


IMS212:
looks good to me!

PR #1339, opened 2022-03-27
Clean up texture swizzle code

IMS212:


src/main/java/net/coderbot/iris/mixin/MixinShaderInstance.java
```diff
@@ -55,9 +50,7 @@
 			return;
 		}
 
-		// TODO: Don't hardcode this list of samplers bound to texture unit 0
-		if (samplerName.equals("Sampler0") || samplerName.equals("tex") || samplerName.equals("texture")
-				|| samplerName.equals("gtexture")) {
+		if (samplerUnit == 0) {
```
coderbot16, line 53: While this addresses a TODO, I'm not much more of a fan of this approach... I think it's worth documenting the fact that this strongly assumes that gtexture always gets bound to texture unit 0, while in the future I'd like to get to the point where this isn't required.

IMS212: Doesn’t it need to be bound to 0, to deal with unrecognized samplers?

coderbot16: That's a driver quirk. I'd like to eventually have code that finds the samplers and figures out which ones are unrecognized instead eventually instead of just relying on driver quirks.

IMS212: Sounds good, will changing to TextureUnit.TERRAIN.unitId fix this for now?

coderbot16: I was only referring to this one file - the rest of the PR is fine.

IMS212: Oh, I’ll remove that then.

IMS212:
Fixed.

IMS212:
uh not sure what happened to git??? fixing

IMS212:
fixed (real)

src/main/java/net/coderbot/iris/pipeline/newshader/ShaderKey.java
```diff
@@ -109,6 +109,10 @@ public boolean isBeaconBeam() {
 		return this == BEACON || this == SHADOW_BEACON_BEAM;
 	}
 
+	public boolean isSwizzle() {
```
coderbot16, line 112: Not needed, see isIntensity

coderbot16 (changes requested):
Looks like this conflicts, also one more comment.

PR #1369, opened 2022-04-17
Update supportedshaders.md

Lolothepro:
https://cdn.discordapp.com/attachments/774352792659820598/965277393143103498/2022-04-17_17.39.57.png
https://cdn.discordapp.com/attachments/774352792659820598/965277392597823599/2022-04-17_17.39.38.png
https://cdn.discordapp.com/attachments/774352792659820598/965277392149049344/2022-04-17_17.38.42.png

It is said that no SEUS works with Mesa, which is not true, as in my experience everything works fine.

### DIFF
--- a/docs/supportedshaders.md
+++ b/docs/supportedshaders.md
@@ -12,9 +12,9 @@
 
 * [Continuum 2.0.4](https://continuum.graphics/downloads)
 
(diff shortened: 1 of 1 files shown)


AlsoSylv:
SEUS renewed and other shaders like it do not work with Mesa. Well it's true a flag fixes this, it breaks other shaders as well, and suggesting users add it does not solve the issue.

PR #1370, opened 2022-04-18
Particle ID system

IMS212:
yee
it's finally happening
debate here


coderbot16:
More of a high-level comment, but I feel like this is far from an ideal abstraction. Making shader pack developers hardcode IDs for specific particles especially when these IDs can change across Minecraft versions will cause extra work for shader pack developers. If it turns out that one of the main uses of this is detecting weather particles, we should have a stronger abstraction for that, instead of forcing devs to hardcode the weather particle IDs.

So far, the two use cases I'm aware of are making some particle emissive, and correctly detecting rain particles. Those seem like things that we could provide a nice abstraction for instead of only having this system.

I don't mean to say that this kind of system doesn't make sense, but I don't think this should be the immediate first approach - ideally, hardcoded ID systems should only be needed as an escape hatch for when the current abstraction misses something - not as the sole and primary way of applying common effects.

I think we should add those rain & emissive abstractions *first*, then see what shader pack devs request, and once we have the main use cases implemented *then* add this system.

The same comments go for the proposed "dimension ID map" system

PR #1374, opened 2022-04-21
Move HorizonRenderer to the pipeline package and don't create it every frame

IMS212:


PepperCode1 (approved):

coderbot16 (approved):
Seems straightforward enough, thanks.

PR #1390, opened 2022-04-27
Block type map support

IMS212:
This adds the custom render layer feature of block.properties.

Uses a Reference2ReferenceOpenHashMap similar to Sodium for the best performance.

### DIFF
--- a/src/main/java/net/coderbot/iris/mixin/MixinItemBlockRenderTypes.java
+++ b/src/main/java/net/coderbot/iris/mixin/MixinItemBlockRenderTypes.java
@@ -1,0 +1,23 @@
+package net.coderbot.iris.mixin;
+
+import net.coderbot.iris.block_rendering.BlockRenderingSettings;
+import net.minecraft.client.renderer.ItemBlockRenderTypes;
+import net.minecraft.client.renderer.RenderType;
+import net.minecraft.world.level.block.state.BlockState;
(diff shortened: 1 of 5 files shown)


src/main/java/net/coderbot/iris/mixin/MixinItemBlockRenderTypes.java
```diff
@@ -0,0 +1,23 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+@Mixin(ItemBlockRenderTypes.class)
+public class MixinItemBlockRenderTypes {
+	@Inject(method = "getChunkRenderType", at = @At("HEAD"), cancellable = true)
+	private static void iris$setCustomRenderType(BlockState arg, CallbackInfoReturnable<RenderType> cir) {
+		if (BlockRenderingSettings.INSTANCE.getBlockTypeIds() != null) {
```
coderbot16, line 16: nit: Should just store `BlockRenderingSettings.INSTANCE.getBlockTypeIds()` in a local variable instead of getting it twice.

coderbot16 (approved):
proof that iris devs favor complementary

PR #1394, opened 2022-04-28
cameraType Iris exclusive uniform

IMS212:
0 - first person/unknown
1 - third person back
2 - third person front

that's it


coderbot16:
Is there any reason to distinguish between third person perspective types rather than this just being a boolean of whether or not the camera is in first person or not?

IMS212:
Not really, I just didn’t see a reason *not to*

coderbot16:
Well, there's mods like freelook that allow the user to move the third person camera around the player model (without rotating the player head/model), instead of having it just look forwards and backwards, and there wouldn't be a coherent way to express that with this approach.

IMS212:
Finally fixed.

coderbot16:
Couldn't we call this `firstPersonCamera` or something?

IMS212:
Sure, I'll also make it a bool.

IMS212:
Uniform is now firstPersonCamera boolean.

coderbot16 (approved):
Code looks good, thanks. This should be added to ShaderDoc as well.

PR #1391, opened 2022-04-28
Use glsl-transformer for shader patching 1.16.x

douira:
This PR adds an abstract system for shader patching through `Patcher` using `glsl-transformer` for program transformation of the shaders. The `Patcher` serves as the central string transformation class while `TransformPatcher` implements this using `glsl-transformer`. Even though 1.16.x doesn't require extensive shader patching with the Triforce patcher, there still is the `AttributeShaderTransformer` which can be implemented using `glsl-transformer`.

A previous PR that was closed to rename the branch: https://github.com/IrisShaders/Iris/pull/1142.
https://github.com/IrisShaders/Iris/pull/1312 was created from https://github.com/IrisShaders/Iris/pull/1311 by cherry-picking the commits actually made to the branch.
The 1.17 PR is https://github.com/IrisShaders/Iris/pull/1312.

This PR targets 1.16.x so that the corresponding 1.17.x and 1.18.x changes may extend upon infrastructure established in this PR.

PR #1395, opened 2022-04-29
Use glsl-transformer for shader patching 1.17.x 

douira:
Shader patching with `glsl-transformer`. Merge https://github.com/IrisShaders/Iris/pull/1391 first.


douira:
Iris won't be supporting 1.17 anymore so this doesn't need to exist anymore.

PR #1400, opened 2022-05-01
Optimize atlas stitching

PepperCode1:
`Stitcher#expand` is overwritten to optimize region creation, thus optimizing the size of the atlas. The block atlas (and some others) is halved in size when using default textures as a result of this change (1024x1024 -> 1024x512). Refer to the javadoc and comments for more information.


IMS212 (approved):
“First time contributor”
idk about that one chief

coderbot16 (commented):
The changes here look good - only thing I would like is a brief comment explaining how the logic in the middle is different from vanilla. Comments along the lines of `// Iris start` and `//Iris end` are valuable as well.

PR #1401, opened 2022-05-01
PBR

IMS212:
yeef
made this to draft
Almost all code is by @PepperCode1


coderbot16:
This PR is very big, and contains many changes unrelated to PBR. I would like to have a preliminary PR that implements things that aren't directly tied to the PBR system like the refactors to macro handling, unification of depth buffer & atlas tracking into a single texture tracker, replacing calls to new Vector3d() with constants, and tweaks to the vertex format code. That would allow us to get those things into the Iris 1.2.5 release so we can test them in prod before putting the rest of PBR through.

coderbot16:
What is the status of this PR? There are a number of issue reports in the #pbr-issue-report channel on GitHub that don't appear to have any follow up or explicit resolution. I'm not fully comfortable reviewing this PR and moving forward with the process of getting it into a release until we can get a list of outstanding issues on the PR & confirmation that the other reported issues are fixed.

IMS212 (commented):
Looks good to me, although blocking on #1599. Can't mark as approved since I made this.

PR #1405, opened 2022-05-01
Update .editorconfig

PepperCode1:
- Add `ij_java_imports_layout`
- Add `trim_trailing_whitespace = true`
- Remove `*.gradle` config

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,22 +1,26 @@
+# http://editorconfig.org
 root = true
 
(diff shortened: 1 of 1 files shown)


coderbot16 (approved):
Thanks!

coderbot16:
Finally we can have a standard import order across all editors instead of every single PR having a different import order convention 🎉

PR #1410, opened 2022-05-06
GPU-based center depth sampling

IMS212:
This uses two textures alongside a simple patched to calculate the center depth on-gpu and smooth it.


src/main/java/net/coderbot/iris/pipeline/patcher/CompositeDepthTransformer.java
```diff
@@ -0,0 +1,28 @@
...
+		}
+
+		if (source.contains("iris_")) {
+			throw new IllegalStateException("Shader is attempting to exploit internal Iris code!");
+		}
+
+		StringTransformations transformations = new StringTransformations(source);
+
+		// replace original declaration (fragile!!! we need glsl-transformer to do this robustly)
+		// if entityColor is not declared as a uniform, we don't make it available
```
coderbot16, line 20: I think you forgot to rename a comment when copy pasting

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
```diff
@@ -927,6 +928,8 @@ public void finalizeLevelRendering() {
 		phase = WorldRenderingPhase.NONE;
 		programStackLog.clear();
 
+		centerDepthSampler.sampleCenterDepth();
```
coderbot16, line 931: We aren't sampling center depth in the right place on any Iris version it seems. I checked ShadersMod and it samples center depth before it renders the hand (which depends on whether or not you're holding something translucent). Though, that has the notable shortcoming that the center depth is never sampled if the hand isn't rendered - we need more robust logic than that.

src/main/java/net/coderbot/iris/postprocess/CompositeRenderer.java
```diff
@@ -234,8 +234,7 @@ private Program createProgram(ProgramSource source, ImmutableSet<Integer> flippe
 		}
 
 		// TODO: Don't duplicate this with FinalPassRenderer
-		// TODO: Parse the value of const float centerDepthSmoothHalflife from the shaderpack's fragment shader configuration
-		builder.uniform1f(UniformUpdateFrequency.PER_FRAME, "centerDepthSmooth", this.centerDepthSampler::getCenterDepthSmoothSample);
+		builder.addDynamicSampler(centerDepthSampler::getCenterDepthTexture, "iris_centerDepthSmooth");
```
coderbot16, line 237: I notice that in this version we don't let gbuffer or shadow programs read the center depth value, as in older versions of Iris. I assume that ShadersMod / OptiFine let shaders do this, likely just returning the value from the previous frame. Do we want to look into that at some point? Doesn't have to be in this PR.

src/main/java/net/coderbot/iris/postprocess/CenterDepthSampler.java
```diff
@@ -1,62 +1,107 @@
...
-	private final RenderTargets renderTargets;
+	private static final double LN2 = Math.log(2);
 	private boolean hasFirstSample;
 	private boolean everRetrieved;
-	private FrameUpdateNotifier fakeNotifier;
+	private final Program program;
+	private final GlFramebuffer framebuffer;
+	private final int texture;
+	private final int altTexture;
+	public CenterDepthSampler(float halfLife) {
```
coderbot16, line 28: please add a blank line in situations like this
```suggestion
	private final int altTexture;

	public CenterDepthSampler(float halfLife) {
```

src/main/java/net/coderbot/iris/postprocess/CenterDepthSampler.java
```diff
@@ -1,62 +1,107 @@
...
-		// TODO: glReadPixels forces a full pipeline stall / flush, and probably isn't too great for performance
-		IrisRenderSystem.readPixels(
-			renderTargets.getCurrentWidth() / 2, renderTargets.getCurrentHeight() / 2, 1, 1,
-			GL11C.GL_DEPTH_COMPONENT, GL11C.GL_FLOAT, depthValue
-		);
+		this.framebuffer.bind();
 
-		return depthValue[0];
+		GlStateManager._bindTexture(altTexture);
+		DepthCopyStrategy.fastest(false).copy(this.framebuffer, texture, null, altTexture, 3, 3);
```
coderbot16, line 79: The API contract of DepthCopyStrategy says it can only copy depth... This will work because the parameter is false, but this deserves a sort of TODO comment addressing this - it should state that it still works even despite the issues.

src/main/java/net/coderbot/iris/postprocess/CenterDepthSampler.java
```diff
@@ -1,62 +1,107 @@
...
+		this.framebuffer.bind();
+		this.program.use();
+
+		RenderSystem.viewport(0, 0, 3, 3);
+
+		FullScreenQuadRenderer.INSTANCE.render();
+
+		GlStateManager._glUseProgram(0);
+
+		GL11C.glDisable(GL11C.GL_SCISSOR_TEST);
```
coderbot16, line 74: I assume this is a vestige of old code?

src/main/java/net/coderbot/iris/postprocess/CenterDepthSampler.java
```diff
@@ -1,62 +1,107 @@
...
+		Minecraft.getInstance().getMainRenderTarget().bindWrite(true);
+	}
+
+	public void setupColorTexture(InternalTextureFormat format) {
+		RenderSystem.texParameter(GL21C.GL_TEXTURE_2D, GL21C.GL_TEXTURE_MIN_FILTER, GL21C.GL_LINEAR);
+		RenderSystem.texParameter(GL21C.GL_TEXTURE_2D, GL21C.GL_TEXTURE_MAG_FILTER, GL21C.GL_LINEAR);
+		RenderSystem.texParameter(GL21C.GL_TEXTURE_2D, GL21C.GL_TEXTURE_WRAP_S, GL21C.GL_CLAMP_TO_EDGE);
+		RenderSystem.texParameter(GL21C.GL_TEXTURE_2D, GL21C.GL_TEXTURE_WRAP_T, GL21C.GL_CLAMP_TO_EDGE);
+
+		GlStateManager._texImage2D(GL21C.GL_TEXTURE_2D, 0, format.getGlFormat(), 3, 3, 0, format.getPixelFormat().getGlFormat(), PixelType.FLOAT.getGlFormat(), null);
```
coderbot16, line 92: Why 3x3 instead of 1x1?

coderbot16 (changes requested):

coderbot16:
This appears to conflict as well

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
IMS212: Not sure how to handle this - we should approach this in a separate PR alongside the next one.

src/main/java/net/coderbot/iris/mixin/MixinRenderTarget.java
```diff
@@ -7,6 +7,7 @@
 import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.ModifyArg;
```
coderbot16, line 10: Spurious import?

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/mixin/MixinRenderTarget.java
IMS212: fixed

coderbot16 (approved):
Great work! Depth of field will now have near-zero performance loss as it always should have had, this will be a big boost for Iris over OptiFine when using DOF packs! 👍 

Could you make separate "TODO issues" for the two things I noted so we remember to fix them later?

PR #1415, opened 2022-05-09
Port non-PBR changes from PBR branch

PepperCode1:
- Unify texture tracking
- Add zero vector constants for uniforms
- Delete `ShaderConstants`
- Add methods to `IrisLogging`, `IrisRenderSystem`, `DepthBufferFormat`, and others
- Fix formatting in some places
- Other miscellanous changes


src/main/java/net/coderbot/iris/mixin/texture/MixinAbstractTexture.java
```diff
@@ -0,0 +1,21 @@
...
+import org.spongepowered.asm.mixin.injection.At.Shift;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+@Mixin(AbstractTexture.class)
+public class MixinAbstractTexture {
+	@Shadow
+	protected int id;
+
+	@Inject(method = "getId()I", at = @At(value = "INVOKE", target = "Lcom/mojang/blaze3d/platform/TextureUtil;generateTextureId()I", shift = Shift.BY, by = 2))
```
coderbot16, line 17: Why not inject at RETURN and get the return value?

coderbot16 (changes requested):
Looks great! Just one small comment.

src/main/java/net/coderbot/iris/mixin/texture/MixinAbstractTexture.java
coderbot16: Fair enough. Maybe we can have a short comment along the lines of:

> Inject after the newly-generated texture ID has been stored into the "id: field

coderbot16 (approved):
Thanks for this, will be nice to have these hopefully known-good changes merged.

PR #1435, opened 2022-05-19
Update checker

IMS212:
... i forgot to make this a pr


IMS212:
The change from IMS212 to IrisShaders should be done after merge, and not during testing. It hosts a testing "update" in English and Spanish.

coderbot16:
Could you put the update checker code into a separate class instead of putting it into Iris.java?

coderbot16:
Also, maybe it would be best to somehow only check for updates once a day or so. People might relaunch the game many times in one day and we likely don't want to send a request on every launch. And of course there's the edge case of someone setting their system clock backwards if that gets implemented.

IMS212:
That would break the “new update detected” message unless we store there’s a new update, which would lead to the issue of what if that update was installed?

coderbot16:
You could cache the full JSON response from the server... That would probably work well enough

IMS212:
Fixed.

src/main/java/net/coderbot/iris/UpdateChecker.java
```diff
@@ -0,0 +1,120 @@
...
+	public UpdateChecker(int simpleVersion) {
+		this.simpleVersion = simpleVersion;
+		if (Objects.equals(System.getProperty("iris.installer", "false"), "true")) {
+			usedIrisInstaller = true;
+		}
+	}
+
+	public void checkForUpdates(IrisConfig irisConfig) {
+		this.info = CompletableFuture.supplyAsync(() -> {
+			if (!irisConfig.shouldDisableUpdateMessage()) {
```
coderbot16, line 46: You can hoist this out of the future, just return from the checkForUpdates method if update checking is disabled.

src/main/java/net/coderbot/iris/UpdateChecker.java
```diff
@@ -0,0 +1,120 @@
...
+				try {
+					File updateFile = FabricLoader.getInstance().getGameDir().resolve("irisUpdateInfo.json").toFile();
+					if (DateUtils.isSameDay(new Date(), new Date(updateFile.lastModified()))) {
+						Iris.logger.warn("Cached update file detected, using that!");
+						UpdateInfo updateInfo = new Gson().fromJson(FileUtils.readFileToString(updateFile, StandardCharsets.UTF_8), UpdateInfo.class);
+						if (updateInfo.simpleVersion > simpleVersion) {
```
coderbot16, line 52: I'm not sure if the simpleVersion stuff is the right approach. This won't show an update warning for someone using `1.2.5-rc.1` if `1.2.5` is the latest veresion. Fabric Loader has some good code for SemVer version comparisons that we probably want to use instead.

src/main/java/net/coderbot/iris/config/IrisConfig.java
```diff
@@ -121,7 +131,8 @@ public void load() throws IOException {
 		properties.load(Files.newInputStream(propertiesPath));
 		shaderPackName = properties.getProperty("shaderPack");
 		enableShaders = !"false".equals(properties.getProperty("enableShaders"));
-		enableDebug = !"false".equals(properties.getProperty("enableDebug"));
+		enableDebug = "true".equals(properties.getProperty("enableDebug"));
```
coderbot16, line 134: You're probably just going to need to rename the option at this point since so many people now have enableDebug=true in their config files from this oversight.

coderbot16 (changes requested):

IMS212:
all your bugs belong to me (Fixed)

src/main/java/net/coderbot/iris/UpdateChecker.java
```diff
@@ -113,8 +129,12 @@ public Component getUpdateMessage() {
 	}
 
 	public String getUpdateLink() {
```
coderbot16, line 131: We're gonna need some `@Nullable` annotations on some of these methods... Or just return Optional

src/main/java/net/coderbot/iris/UpdateChecker.java
```diff
@@ -29,58 +32,71 @@
...
 
-					try (InputStream in = new URL("https://raw.githubusercontent.com/IMS212/Iris-Installer-Files/master/updateindex.json").openStream()) {
-						String updateIndex = new JsonParser().parse(new InputStreamReader(in)).getAsJsonObject().get(StandardMacros.getMcVersion()).getAsString();
-						String json = IOUtils.toString(new URL(updateIndex), StandardCharsets.UTF_8);
-						UpdateInfo updateInfo = new Gson().fromJson(json, UpdateInfo.class);
-						BufferedWriter writer = new BufferedWriter(new FileWriter(updateFile));
-						writer.write(json);
-						writer.close();
-						if (updateInfo.simpleVersion > simpleVersion) {
+				try (InputStream in = new URL("https://raw.githubusercontent.com/IMS212/Iris-Installer-Files/master/updateindex.json").openStream()) {
```
coderbot16, line 72: I think the content of the files at these URLs should probably be changed before we put anything out for testing... also, it might be best to put the index on GitHub releases somewhere like what we do for the other installer files. Though technically these aren't really installer files :P

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/UpdateChecker.java
IMS212: Putting it on github releases probably isn't smart, as it could cause issues in the future with indexing, especially if a ISP tries caching "latest" releases. Will change the content.

coderbot16: But that's exactly what we do for the installer already, and it seems to work just fine.

IMS212:
Changed messages and changed to use Optional.

src/main/java/net/coderbot/iris/UpdateChecker.java
IMS212: We may want to update an index with more revisions down the line to accomodate changes, using Github Releases is bad for that. Github Releases shouldn't be used for text that has the possibility of being changed.

coderbot16: In that case you just make a new release. This has been working for nearly a year with the Iris Installer - it can easily pull from `https://github.com/IrisShaders/Iris-Installer-Files/releases/latest`.

IMS212: I guess that works; I'll try it.

IMS212: Done.

src/main/java/net/coderbot/iris/mixin/MixinClientPacketListener.java
```diff
@@ -17,8 +17,8 @@ public class MixinClientPacketListener {
 
 	@Inject(method = "handleLogin", at = @At("TAIL"))
 	private void iris$showUpdateMessage(ClientboundLoginPacket arg, CallbackInfo ci) {
-		if (this.minecraft.player != null && Iris.getUpdateChecker().getUpdateMessage() != null) {
-			this.minecraft.player.displayClientMessage(Iris.getUpdateChecker().getUpdateMessage(), false);
+		if (this.minecraft.player != null && Iris.getUpdateChecker().getUpdateMessage().isPresent()) {
+			this.minecraft.player.displayClientMessage(Iris.getUpdateChecker().getUpdateMessage().get(), false);
```
coderbot16, line 21: Ideally we'd use ifPresent here.

coderbot16 (approved):
Looks good for now 👍

PR #1438, opened 2022-05-19
Add transform patcher base to trunk

douira:
This PR adds `TransformPatcher` without actually using it to do anything. Patching for various things can then be added on top of this. Any patching happens in `TransformPatcher` and new methods are added to it and `Patcher` when new types of things should be patched.


douira:
closing this for now. this is example code that can be used elsewhere. I'm making a new PR for incremental adoption.

PR #1443, opened 2022-05-20
Slight extended vertex data refactor

PepperCode1:
- Combine vanilla and Sodium NormalHelper and QuadView
- Move tangent calculation to NormalHelper
- Make QuadView an interface and split into PolygonView, TriView, and
QuadView
- Make Sodium entity writers more similar to Sodium terrain writers
- Explicitly check vertex mode before filling extended data in
MixinBufferBuilder
- Support extended data for tris in MixinBufferBuilder
- Convert custom texture id maps to EnumMaps


src/main/java/net/coderbot/iris/mixin/vertices/MixinBufferBuilder.java
```diff
@@ -172,121 +142,47 @@ public void endBlock() {
...
-		float dot = (bitangentx * pbitangentx) + (bitangenty * pbitangenty) + (bitangentz * pbitangentz);
-		float tangentW;
-
-		if (dot < 0) {
-			tangentW = -1.0F;
+		midU /= vertexAmount;
+		midV /= vertexAmount;
+
+		if (vertexAmount == 3) {
+			NormalHelper.computeFaceNormal(normal, (TriView) polygon);
```
coderbot16, line 168: I like most of the code here but the fact that this changes behavior just based on a seemingly unnecessary cast seems a bit too implicit for me. I don't want removing a seemingly unnecessary cast to change the behavior of the program.

PepperCode1: Possible solutions:
- Rename the method to `computeFaceNormalTri` or similar
- Split `BufferBuilderPolygonView` into `BufferBuilderTriView` and `BufferBuilderQuadView` and apply one of the following:
  - Allow the polygon itself or a separate handler to fill in the vertex data
  - Duplicate the `fillExtendedData` method (one for quads, one for tris)

Which solution should be applied?

coderbot16: The rename approach seems to be the most straightforward for now.

coderbot16 (approved):
Looks good, thanks.

PR #1451, opened 2022-05-22
(In 1.18.2) For compatibility with Immersive Portals: Use version number to track whether the color buffer and depth buffer should refresh.

qouteall:
I roughly tested it and the window resizing works both with and without immptl.


maximumpower55:
This PR isn't needed anymore the one for trunk has been merged and ported

PR #1456, opened 2022-05-23
For compatibility with Immersive Portals: Use version number to track whether the color buffer and depth buffer should refresh

qouteall:
https://github.com/IrisShaders/Iris/pull/1451 but in trunk branch.


coderbot16 (approved):
Looks great, thanks!

PR #1464, opened 2022-05-27
Copy initial depth texture to any extra depth textures to ensure data is not invalid

IMS212:


IMS212:
Will merge tomorrow if there are no complaints as this is marked high priority.

PR #1465, opened 2022-05-28
Use addDepthStencilAttachment instead of addDepthAttachment if the framebuffer has stencil.

qouteall:
Without this, stencil buffer cannot really work with shaders on.


IMS212 (approved):
Looks good to me.

coderbot16 (approved):
Thanks, nice catch.

maximumpower55 (approved):
:b:erge

BeansBeefBroccoli:
Yes

PR #1466, opened 2022-05-30
Text sink API

IMS212:


src/main/java/net/irisshaders/iris/api/v0/IrisApi.java
```diff
@@ -22,7 +25,7 @@ static IrisApi getInstance() {
 	 * if they wish to check whether given API calls are available on
 	 * the currently installed Iris version.
 	 *
-	 * @return The current minor revision. Currently, revision 0.
+	 * @return The current minor revision. Currently, revision 1.
```
SquidDev, line 28: This still seems to return `0` in the impl.

src/main/java/net/coderbot/iris/vertices/IrisTextVertexSinkImpl.java
```diff
@@ -0,0 +1,89 @@
...
+	private final ByteBuffer buffer;
+	private final QuadViewEntity.QuadViewEntityUnsafe quad = new QuadViewEntity.QuadViewEntityUnsafe();
+	private final Vector3f saveNormal = new Vector3f();
+	private static final int STRIDE = IrisVertexFormats.TERRAIN.getVertexSize();
+	private int vertexCount;
+	private long elementOffset;
+	private float uSum;
+	private float vSum;
+
+	public IrisTextVertexSinkImpl(int maxQuadSize, IntFunction<ByteBuffer> buffer) {
```
coderbot16, line 23:
```suggestion
	public IrisTextVertexSinkImpl(int maxQuadCount, IntFunction<ByteBuffer> buffer) {
```

src/main/java/net/coderbot/iris/vertices/IrisTextVertexSinkImpl.java
```diff
@@ -0,0 +1,89 @@
...
+			for (long vertex = 0; vertex < 4; vertex++) {
+				MemoryUtil.memPutFloat(i + 36 - STRIDE * vertex, uSum);
+				MemoryUtil.memPutFloat(i + 40 - STRIDE * vertex, vSum);
+				MemoryUtil.memPutInt(i + 28 - STRIDE * vertex, normal);
+				MemoryUtil.memPutInt(i + 44 - STRIDE * vertex, tangent);
+			}
+
+			uSum = 0;
+			vSum = 0;
+		}
```
coderbot16, line 84: We can actually compute these values at the head of `quad()` instead of after all vertices are written, and can have some neat optimizations based on the limits of the function call. I understand if you don't want to implement any of that now, but we could have a TODO comment for that still.

src/main/java/net/irisshaders/iris/api/v0/IrisTextVertexSink.java
```diff
@@ -0,0 +1,11 @@
...
+
+import com.mojang.blaze3d.vertex.VertexFormat;
+
+import java.nio.ByteBuffer;
+
+public interface IrisTextVertexSink {
+	VertexFormat getUnderlyingVertexFormat();
+	ByteBuffer getUnderlyingByteBuffer();
+	void quad(float x1, float y1, float x2, float y2, float z, int rgba, float u1, float v1, float u2, float v2, int light);
+}
```
coderbot16, line 11: We need javadoc comments here

src/main/java/net/irisshaders/iris/api/v0/IrisApi.java
```diff
@@ -94,4 +97,10 @@ static IrisApi getInstance() {
 	 * @since API v0.0
 	 */
 	IrisApiConfig getConfig();
+
+	/**
+	 * Gets a text vertex sink to render into.
+	 * @since API 0.1
+	 */
+	IrisTextVertexSink createTextVertexSink(int maxQuadCount, IntFunction<ByteBuffer> bufferProvider);
```
coderbot16, line 105: We should document the semantics of bufferProvider - how ownership works, etc

src/main/java/net/irisshaders/iris/api/v0/IrisTextVertexSink.java
```diff
@@ -0,0 +1,34 @@
...
+import java.nio.ByteBuffer;
+
+public interface IrisTextVertexSink {
+	/**
+	 * Gets the underlying vertex format used for rendering text.
+	 * @return a valid {@code VertexFormat} instance
+	 */
+	VertexFormat getUnderlyingVertexFormat();
+	/**
+	 * Gets the underlying vertex format used for rendering text in the current sink.
```
coderbot16, line 14:
```suggestion
	 * Gets the underlying buffer used for rendering text in the current sink.
```

coderbot16 (approved):
Nice work, thansk!

PR #1488, opened 2022-06-14
Use version counter to check whether to reload sodium shaders, for compatibility with Immersive Portals

qouteall:
Fix https://github.com/IrisShaders/Iris/issues/1188

Immersive Portals' custom dimension also use the overworld pipeline. However, every dimension have one `LevelRenderer` and one `IrisChunkProgramOverrides`. It only set `sodiumShaderReloadNeeded` when creating a new pipeline, so the shader in custom dimension won't be reloaded. This issue does not manifest without ImmPtl because only one client world is present. This PR fixes this by using version counter to determine whether to reload the sodium shader.

Also removes unused `PipelineManager.getInstance()` (It uses `Iris.getPipelineManager()`).

How to solve merge confilct in 1.18.2 branch: https://github.com/qouteall/Iris/commit/5bc76294e515d21ce768ae508e81dc7162a663f5

### DIFF
--- a/src/main/java/net/coderbot/iris/pipeline/PipelineManager.java
+++ b/src/main/java/net/coderbot/iris/pipeline/PipelineManager.java
@@ -1,8 +1,11 @@
 package net.coderbot.iris.pipeline;
 
 import com.mojang.blaze3d.platform.GlStateManager;
+import me.jellysquid.mods.sodium.client.gl.device.RenderDevice;
+import me.jellysquid.mods.sodium.client.render.chunk.passes.BlockRenderPass;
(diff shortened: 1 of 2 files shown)


IMS212:
This seems moreso a bandaid fix than anything, but it does work. I'll wait for a proper review.

coderbot16:
I think this is a good change, but there should be at least one code comment explaining why a version number is necessary. You had a detailed writeup in the comment on the issue and I think that if we had that in the PR it would be nice.

qouteall:
Comment added.

PR #1498, opened 2022-06-18
Korean translation improvement

xphere07:
Current Korean translation has old-version-text and some wrong orthography. It's needed to be fixed.

Note: `쉐이더` is commonly used **wrong** notation of the English word `shader`. It must be fixed into `셰이더`.


IMS212:
LGTM, thanks!

PR #1508, opened 2022-06-23
Rename Uniform Prefixes to iris_ for better future integration with glsl-transformer

douira:
This change only renames the renamable uniforms to use `iris_` instead of `u_`, `d_` or `a_` as a prefix. This PR does not introduce glsl-transformer but it makes it easier to work with it in its own PR.


IMS212:
Yes and no, it’ll require extra patches to Sodium.

PR #1523, opened 2022-07-06
Begin pass

IMS212:
A composite pass that runs right after render target setup

### DIFF
--- a/src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
+++ b/src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
@@ -88,7 +88,7 @@ public class DeferredWorldRenderingPipeline implements WorldRenderingPipeline, R
 	private final ImmutableList<ClearPass> clearPasses;
 
 	private final GlFramebuffer baseline;
-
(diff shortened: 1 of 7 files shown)


src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
```diff
@@ -186,7 +184,7 @@ public DeferredWorldRenderingPipeline(ProgramSet programs) {
 
 		GlStateManager._activeTexture(GL20C.GL_TEXTURE0);
 
-		this.flippedBeforeShadow = ImmutableSet.of();
+		this.flippedAfterBegin = ImmutableSet.of();
```
coderbot16, line 187: This seems to either be wrong or unnecessary given the assignment below.

coderbot16 (changes requested):

src/main/java/net/coderbot/iris/pipeline/DeferredWorldRenderingPipeline.java
IMS212: IIRC compiler complains otherwise, I’ll check later.

PR #1543, opened 2022-07-16
Unset tab_width on types with indent_style = tab

douira:
If tabs are used, the indent width is purely visual and should not be set by the editorconfig and has no effect on the files themselves.


IMS212:
ok fine

PR #1545, opened 2022-07-17
Per-buffer blending

IMS212:
yeah


src/main/java/net/coderbot/iris/shaderpack/ShaderProperties.java
```diff
@@ -192,8 +198,44 @@ public ShaderProperties(String contents, ShaderPackOptions shaderPackOptions, It
...
+						bufferBlendOverrides.computeIfAbsent(parts[0], list -> new ArrayList<>()).add(new BufferBlendOverride(index, null));
+						return;
+					}
+
+					String[] modeArray = value.split(" ");
+					int[] modes = new int[4];
+
+					int i = 0;
+					for (String modeName : modeArray) {
+						modes[i] = BlendModeFunction.fromString(modeName).get().getGlId();
```
MahirZukic, line 233: We could get an `ArrayIndexOutOfBounds` Exception here due to the 229th line.
Shouldn't that be?
`int[] modes = new int[modeArray.length];`

MahirZukic (changes requested):
We could get an ArrayIndexOutOfBounds Exception like I already commented.

src/main/java/net/coderbot/iris/shaderpack/ShaderProperties.java
IMS212: Thanks for pointing this out, fixed

IMS212:

PR #1547, opened 2022-07-20
mainUnitSize uniform, atlasSize for non-atlases

IMS212:


coderbot16:
I would have no idea what `mainUnitSize` meant without looking at the code. Maybe gtextureSize or something like that instead?

IMS212:
Done.

coderbot16:
Why not just make this unconditionally available? I don't think it's necessary to make it only available for non-atlases.

IMS212:
I guess so, will do.

IMS212:
done

IMS212:

coderbot16 (approved):
Looks good!

PR #1548, opened 2022-07-20
Update sk_sk.json

SmajloSlovakian:
i have made this mistake longer ago and also fixed it not long after that... now i can't see any pull request by me... strange

i also added some new strings


IMS212:
Thanks!

PR #1550, opened 2022-07-22
preserve gui render state when vignette is disabled

Sorenon:
Just a small change to vignette disabling to correctly set up the render state for the GUI
This was causing a bug in MCXR where the hotbar was being drawn incorrectly

This better matches how vanilla does it:
```
if (Minecraft.useFancyGraphics()) {
    this.renderVignette(this.minecraft.getCameraEntity());
} else {
    RenderSystem.enableDepthTest();
    RenderSystem.defaultBlendFunc();
 }
```


IMS212:
Looks good, thanks!

PR #1561, opened 2022-07-25
Add Optifine SkyBox support

EpicGamer121:
Add Support for skybox texturepacks that use optifine


IMS212:
“no”, he said, as he closed the PR

PR #1563, opened 2022-07-26
Feature: proof of concept custom item lighting

parzivail:
Through the use of an interface that mods can implement, items can provide custom lighting values that vary based on the target player and stack. This augments the existing functionality of stack-based item lighting which was limited to blocks.

### DIFF
--- a/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
@@ -258,12 +258,11 @@ public final class CommonUniforms {
 
 			ItemStack stack = client.player.getItemInHand(hand);
 
-			if (stack == ItemStack.EMPTY || stack == null || !(stack.getItem() instanceof IrisItemLightProvider)) {
+			if (stack == ItemStack.EMPTY || stack == null) {
 				return 0;
 			}
 
-			IrisItemLightProvider item = (IrisItemLightProvider)stack.getItem();
-			return item.getLightEmission(client.player, stack);
+			return ((IrisItemLightProvider)stack.getItem()).getLightEmission(client.player, stack);
 		}
 	}
 
--- a/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
+++ b/src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
@@ -33,7 +33,6 @@ import net.minecraft.world.effect.MobEffectInstance;
 import net.minecraft.world.effect.MobEffects;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.LivingEntity;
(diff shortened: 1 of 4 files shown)


coderbot16:
You'd probably want to implement this on `Item` in a Mixin so that mods don't need to have the `implements IrisItemLightProvider` clause on their own item classes - which would lead to a hard dependency on Iris.

parzivail:
I recognize that the interface being unavailable at runtime will generate an exception, this is just to get the ball rolling

parzivail:
@coderbot16 oops, was writing that comment and didn't see yours: that's a great idea. I'll get to work on it

parzivail:
How okay would you be with me exposing a new pair of uniforms for mods and shader authors to take advantage of, as a part of this API, that allow items to also optionally specify which color they wish to emit?

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -259,22 +258,12 @@ public int getAsInt() {
...
-			}
-			else if (stack.getItem() instanceof IrisItemLightProvider) {
-				IrisItemLightProvider item = (IrisItemLightProvider)stack.getItem();
-
-				return item.getLightEmission(client.player, stack);
-			}
-
-			return 0;
+			IrisItemLightProvider item = (IrisItemLightProvider)stack.getItem();
+			return item.getLightEmission(client.player, stack);
```
maximumpower55, line 266:
```suggestion
			return ((IrisItemLightProvider)stack.getItem()).getLightEmission(client.player, stack);
```

Remember to remove the line above!

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
```diff
@@ -259,22 +258,12 @@ public int getAsInt() {
 
 			ItemStack stack = client.player.getItemInHand(hand);
 
-			if (stack == ItemStack.EMPTY || stack == null) {
+			if (stack == ItemStack.EMPTY || stack == null || !(stack.getItem() instanceof IrisItemLightProvider)) {
```
maximumpower55, line 261: > || !(stack.getItem() instanceof IrisItemLightProvider)
This part of this if statement is useless, Items will always have a implementation of `IrisItemLightProvider` because of the mixin.

maximumpower55 (commented):
Looks good! expect for a couple nitpicks

src/main/java/net/coderbot/iris/uniforms/CommonUniforms.java
parzivail: Good point! Slipped my mind.

maximumpower55 (approved):
LGTM!

PR #1575, opened 2022-07-31
Disable cloud settings with Sodium if it's being overriden

IMS212:
This doesn't work on Vanilla yet.


coderbot16:
I assume we'd want an implementation for vanilla as well before merging? While it might be impractical I'd like for Iris to always work without Sodium.

IMS212:
Yes, however it’ll be quite hard due to how the positioning of options are laid out

PR #1579, opened 2022-08-03
Buffer size support

IMS212:
Needs testing.


IMS212:

PR #1585, opened 2022-08-08
Remove all Fabric API dependencies

IMS212:
Will make ports to other platforms much much easier, alongside removing dependencies.


src/main/java/net/coderbot/iris/mixin/MixinMinecraft_Images.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * This Mixin is responsible for registering the "widgets" texture used in Iris' GUI's.
+ * Normally Fabric API would do this automatically, but we don't use it here, so it must be done manually.
+ */
+@Mixin(Minecraft.class)
+public class MixinMinecraft_Images {
+	@Inject(method = "<init>", at = @At("TAIL"))
+	private void iris$setupImages(GameConfig arg, CallbackInfo ci) {
+		try {
+			Minecraft.getInstance().getTextureManager().register(new ResourceLocation("iris", "textures/gui/widgets.png"), new NativeImageBackedCustomTexture(new CustomTextureData.PngData(new TextureFilteringData(false, false), IOUtils.toByteArray(Iris.class.getResourceAsStream("/assets/iris/textures/gui/widgets.png")))));
```
coderbot16, line 27: Have you tested that this works properly after a resource reload? I'm concerned that these kinds of textures will be deleted during a reload.

IMS212: Just tested it, works fine.

IMS212:

PR #1589, opened 2022-08-10
Entity ID and block entity ID attribute transformations

IMS212:
This uses patching to replace existing defines of entityId and blockEntityId with attribute variants and injects those attributes. This is required for future instanced rendering, alongside not stalling the GPU with uniforms.


coderbot16:
This is outdated and needs an update for glsl-transformer. Also, is there any particular reason to re-use the mc_Entity attribute name instead of using our own attribute name?

douira:
I believe we have in the mean time also changed a lot of the code this is referencing.

IMS212:
yep, closing for now

PR #1594, opened 2022-08-13
Rename ua_ua.json to uk_ua.json

Borusu1:
In the file, the translation is in Ukrainian, but it is not called uk_ua


coderbot16 (approved):
Thanks! The wiki confirms that you are indeed correct: https://minecraft.fandom.com/wiki/Language

PR #1596, opened 2022-08-18
Add a property to ignore the installed Sodium version

colbiedison:
When building Sodium from source, Iris refuses to load because it does not recognize the Sodium build as compatible, even if it is. I've added a property `-Diris.allowAnySodiumVersion=true` to skip the version check and allow any version of Sodium, for those who know what they're doing.

### DIFF
--- a/src/main/java/net/coderbot/iris/compat/sodium/SodiumVersionCheck.java
+++ b/src/main/java/net/coderbot/iris/compat/sodium/SodiumVersionCheck.java
@@ -28,6 +28,9 @@ public class SodiumVersionCheck {
 	}
 
 	public static boolean isAllowedVersion(String sodiumVersion) {
+		if (Boolean.parseBoolean(System.getProperty("iris.allowAnySodiumVersion")))
+			return true;
+
 		for (AllowedSodiumVersion allowed : ALLOWED_SODIUM_VERSIONS) {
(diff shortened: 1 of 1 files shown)


IMS212:
I feel like this could be abused quite easily, for example if a Youtube tutorial starts "recommending" this to bypass checks.

I don't think it's smart to put this in production.

coderbot16:
I agree with IMS - this seems like a development tool, not something that should be enabled for production usage, especially since the symptoms of an incompatible Sodium version aren't always immediately obvious. Sodium and Iris versions are tightly-coupled (and Sodium has never offered any sort of stable API to Iris - Iris is tightly integrated deep into Sodium internals), and without a baseline level of diagnostics indicating that this option was passed, this could be very problematic for our support team. I'm not really seeing the user need that this option would solve, and so I'd stick to the status quo of strongly-coupled Sodium versions outside of development environments. If it is indeed the case that this solves a direct user need that cannot be solved otherwise, then this might deserve further consideration, but I have not yet seen that.

In addition, if there are specific versions of Sodium that you believe to be compatible that should be marked as so, they could also be added. But this is just too large of an escape hatch to be added as-is. I never like closing a PR from a first-time contributor, but in this case I don't see many other options. But thanks for your interest in contributing!

PR #1611, opened 2022-08-28
gbuffers_text

IMS212:
does what it says


coderbot16 (commented):
I think there's a valid need here, but I'm not sure if this PR dives deeply enough into the issue. What's so special about text that it needs a separate program? Is it the vertex format? Is it the fact that it's always drawn in 2D? Isn't this already a misnomer due to being shared with maps? Could this be addressed instead by signalling to the shader that PBR attributes aren't available somehow?

IMS212:
Closing this while we consider alternate solutions.

PR #1619, opened 2022-08-31
Shadow render targets rewrite

IMS212:
This is much closer to how normal render targets work.


IMS212:
This has been included in the latest Discord beta. If no issues are found, this will be merged.

It was also discovered this fixes the "unknown performance loss" with Optimus drivers.

PR #1645, opened 2022-09-24
gbuffers_entities_trans

IMS212:
people have been asking for this for literally 5 years and I added it in 20 minutes

gg optifine


coderbot16:
I don't think this is robust enough... This will actually need the blending state to be checked instead of hardcoding render type names.

IMS212:
Sounds good, will do. Is the program usage itself OK?

coderbot16:
The setup looks fine but I think this is something that should be checking the blend state from glstatemanager instead of involving itself with render types at all.

IMS212:
hm, ok. That works, I guess.

IMS212:
Fixed.

IMS212:
Will merge soon. This optionally depends on feature flags existing.

Issue #1229, opened 2022-01-20
"Hide GUI" Button


Turn999:
A button somewhere in the Corner where you can hide the sliders, buttons and borders so you can view the scenery in your current position without going out of the settings menu. Just to make it easier to view the changes without always having to navigate back to shader settings whenever you want to make additional small adjustments.

OttCS:
This would be a great feature! For the mechanic, I'd support "hiding the settings" as long as you're holding down a button (tab, shift, etc.) and they come back up when you let go.

I'm not sure about the feasibility of this, the esc key is usually the only thing used. A button that goes to a "blank screen" and can be exited with the esc key would be good inside the shader settings page (the shader settings background is grayed out AND has buttons all over it, a little tough to see through).

douira:
This issue can probably be closed now since #1657 implemented this feature.

PR #1657, opened 2022-10-01
Add a button to hide the settings GUI

ConsoleLogLuke:
This simply adds a button to the settings screen that hides or unhides the GUI, as described in #1229.

The X position of the button is dynamic:
- If the screen is large, it'll be 50 units away from the right edge of the screen
- If the screen is too small to fit the button, it'll be on the very right edge of the screen
- If the screen is otherwise small, it'll be centered within the free space on the right side of the screen

The button has different textures depending on if the GUI is hidden or not, and it displays a tooltip when you hover over it for long enough. The button itself and the "Iris [version]" text still remain shown when the GUI is hidden. You can hide the GUI either when you're selecting or you're configuring a shader pack.

Normal GUI
Hidden GUI


IMS212:
Looks good to me! Will merge soon if no other developers have issues with it.

PepperCode1 (approved):

PR #1667, opened 2022-10-02
Implement item light color providing

parzivail:
* Much like item light providing, this solution relies on an interface mixin that provides a default method.
* Color can be provided by an item each frame. I originally thought each tick to save performance, but well-optimized mods could take advantage of per-frame color.
* Implements uniforms for the main and offhand using a 3D float vector instead of a 4D one, since I don't know what transparency would mean in this context
* Will require shader support/implementation before the feature can be realized in clients


src/main/java/net/irisshaders/iris/api/v0/item/IrisItemColorProvider.java
```diff
@@ -0,0 +1,14 @@
+package net.irisshaders.iris.api.v0.item;
+
+import net.coderbot.iris.vendored.joml.Vector3f;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+
+public interface IrisItemColorProvider
```
maximumpower55, line 7: This shouldn't be a separate interface Imo

src/main/java/net/irisshaders/iris/api/v0/item/IrisItemColorProvider.java
```diff
@@ -0,0 +1,14 @@
...
+{
```
maximumpower55, line 8: Why is the `{` on a separate line, this isn't csharp lmao

maximumpower55 (approved):
LGTM, just some minor nitpicks

src/main/java/net/irisshaders/iris/api/v0/item/IrisItemColorProvider.java
parzivail: Sure, I can merge it into `IrisItemLightProvider` if it's a better fit

src/main/java/net/irisshaders/iris/api/v0/item/IrisItemColorProvider.java
parzivail: oops, IDE settings

src/main/java/net/coderbot/iris/mixin/MixinItem.java
```diff
@@ -1,9 +1,11 @@
...
 
+import net.irisshaders.iris.api.v0.item.IrisItemColorProvider;
 import net.irisshaders.iris.api.v0.item.IrisItemLightProvider;
 import net.minecraft.world.item.Item;
 import org.spongepowered.asm.mixin.Mixin;
 
 @Mixin(Item.class)
-public class MixinItem implements IrisItemLightProvider {
+public class MixinItem implements IrisItemLightProvider, IrisItemColorProvider
+{
```
maximumpower55, line 10: Same here why is the `{` on a separate line

src/main/java/net/irisshaders/iris/api/v0/item/IrisItemColorProvider.java
IMS212: Not sure if this should use JOML's Vector3f. This makes it hard for other mods to support without a second dependency, while MC's Vec3 would work fine.

PR #1672, opened 2022-10-04
Add compat for item lighting provided through LambDynamicLights

parzivail:
Adds "free" support for mods which register dynamic item lighting through LambDynamicLights via a compat module


IMS212 (changes requested):
I'm not comfortable at all with this reflection, especially if it's done every uniform update.

parzivail:
[Discord discussion resulted in compat changing hands to LambDynamicLights instead]

PR #1684, opened 2022-10-07
Alphabetize incompatibilities

DamianRivas:
I found it just a tad bit tedious to see if my massive modpack had incompatibilities so I reordered the list to be alphabetized


IMS212:
Thanks!

PR #1687, opened 2022-10-07
new: ExtendedShader activation system similar to 1.16

IMS212:
Testing will be done in the 1.5 cycle.


IMS212:
Merging, as it's required for many things in the 1.5 cycle. Any bugs with this should be caught quickly in betas, as it's required for any shader to work.

PR #1688, opened 2022-10-09
Update Korean translation

xphere07:
Updating Korean translations by filling missed contexts and fixed some punctuation marks to fit Korean orthography.


IMS212:
Thanks!

PR #1704, opened 2022-10-15
Update Japanese translation

KabanFriends:
Changes:
- Added missing entires in `ja_jp.json`
- Changed existing translations to be consistent with Minecraft's Japanese translations
- Fixed incorrect characters in %s placeholders (Fixes the following issue)
image


IMS212:
Thanks!

PR #1714, opened 2022-10-23
Fix many resource leaks

altrisi:
This pull request fixes _many_ resource (not memory) leaks in the Iris codebase. Most common ones are not closing the streams from `Files.walk` and `Files.list`, needed because `Stream`s don't run their closing operations when you run their terminal operation. Fun thing is there were some handled correctly around.

There were also some Input/OutputStreams that were never closed and are now. I've looked at the whole codebase and tried to fix all of the resource leaks (that I know) related to the `Files` class usage.

After this (if I tested it right), you can delete shaderpacks files without having to stop the game.


maximumpower55 (approved):
Code looks good to me

IMS212:
Thanks!

PR #1749, opened 2022-11-25
Debug screens (1.19.3 only)

IMS212:


douira:
It seems IrisConfig now also does more than just storing the path to the current shaderpack, as the class' javadoc falsely claims. Is there potential to store more nuanced debugging options in this too? The shaderpack printer and some other possible debug features have a number of configurable options. (transformer print type, patched_shaders printing, possibly increasing patching verbosity etc)

IMS212:
Yes, it stores all persistent options, more options can be added on request.

PR #1767, opened 2022-12-04
Update Korean translation

xphere07:
Updating Korean translation for lately added strings and general translation improvement.


IMS212:
Thanks!

PR #1791, opened 2022-12-14
Entity ID and block entity ID attribute transformations (v2)

IMS212:
This uses patching to replace existing defines of entityId and blockEntityId with attribute variants and injects those attributes. This is required for future instanced rendering, alongside not stalling the GPU with uniforms.

This is version 2 of this PR, with GLSL transformer support and better patching.


src/main/java/net/coderbot/iris/gl/shader/ProgramCreator.java
```diff
@@ -18,6 +18,7 @@ public static int create(String name, GlShader... shaders) {
 
 		// TODO: This is *really* hardcoded, we need to refactor this to support external calls
 		// to glBindAttribLocation
+		GlStateManager._glBindAttribLocation(program, 11, "iris_Entity");
 		GlStateManager._glBindAttribLocation(program, 11, "mc_Entity");
```
maximumpower55, line 22: It don't think `iris_Entity` and `mc_Entity` should both be bound at `11`

IMS212: We need to preserve attributes locations.

maximumpower55: Wouldn't they conflict if they were both bound at the same location?
or am I missing something here

maximumpower55: Ah

coderbot16: > We need to preserve attributes locations.

Why? Shaders shouldn't be dependent on this information at all. If they are we should definitely be patching them to not be dependent.

IMS212: By preserving locations, I mean we only have 16 of them. We should be reusing when possible.

PR #1885, opened 2023-02-14
Iris 1.6

IMS212:


douira:
patcher related things look good to me. We can defer the refactoring of the `PatchShaderType` things to a later time. (we can discuss what can be done)

PR #1891, opened 2023-02-15
Improve number parsing for settings and consts

xirreal:


coderbot16:
Look at the code again, I wonder if it's a better approach rework the parsing? First by splitting the string on the first `//` occurrence and then looking for whitespace or similar in the value portion.

I'm worried that the process of detecting the end of the float is still just making a weak approximation of actual Float parsing, so if there's some way to avoid that that would be preferable. Since this method might break if we have something like this:

```
#define OPTION 1.0//oops [1.0 2.0]
```

xirreal:
A better method would be preferrable, instead of doing everything char by char, splitting by whitespace and trimming makes much more sense imo.

PR #1923, opened 2023-03-23
Full Optifine core specification support

IMS212:


douira (changes requested):
See https://github.com/IrisShaders/Iris/pull/1924 for suggested changes.

PR #1943, opened 2023-04-12
Documentation on unsized array specifier move

douira:


douira:
Included by https://github.com/IrisShaders/Iris/pull/1958

PR #1951, opened 2023-04-17
Drop of CF

lumaa-dev:
Iris and Sodium are no longer available on CurseForge


Felix14-v2:
They are, so use [Modrinth](https://Modrinth.com) or [IrisShaders.net](https://IrisShaders.net)

PR #1955, opened 2023-04-20
Contextual self-promo

Madis0:
Because getting mods one by one can be overwhelming indeed


IMS212:
This will be closed, as these modpacks don't replicate Optifine and instead try to act as "what they wish Optifine had". We'll think of alternate things to put there.

PR #1985, opened 2023-05-06
Fix Traditional Chinese localization

notlin4:


IMS212:
Thanks, however could you rebase this on 1.18.2?

PR #2029, opened 2023-06-16
Fixed image formatting

Greymagic27:
This PR fixes some image formatting in the installation guide page. No steps have been removed, however they now display properly


topia:
It seems to be a well-known side-effect of deleting a target branch (in this case, the branch was recreated after deleting); I believe no changes were applied.

Greymagic27:
Ugh. I’ll remake this then I guess

PR #2056, opened 2023-07-16
Lang (de_de): Fix grammar mistake

yamayaki1:
Even though there is only one alternative mod mentioned for NEC since b6686b39b3fae05636f5aa647a34868fb3caa180, the translation string is still written in the plural form, which makes no sense in this context.


douira:
Lgtm

PR #2075, opened 2023-08-05
Update Estonian

Madis0:
On the correct branch:tm:


IMS212:
This is still not the correct branch, the correct branch is 1.18.2. I'll cherrypick it.

Issue #2113, opened 2023-08-31
GBuffer blend mode "leaks" into the post-processing pipeline


moyongxin:
### What happened?

In the case of sixthsurge/photon#33:
```
// shaders.properties
blend.gbuffers_hand_water.colortex0            = SRC_ALPHA ONE_MINUS_SRC_ALPHA ONE ZERO
```
When this line is commented, the problem seems to be fixed.
I guess that the blend mode "leaks" into the post-processing pipeline so that the first frame after switching to stained glasses gets pure white.

### Minecraft Version

Minecraft 1.20.1

### Iris Version

iris-mc1.20.1-1.6.8.jar

### Operating System

Windows 11

### What is your GPU?

Nvidia Geforce RTX 3050 Laptop

GeForceLegend:
`net.coderbot.iris.gl.IrisRenderSystem.enableBufferBlend()` will enable glBlend without notify GlStateManager, and when calling `RenderSystem.disableBlend()` before composite program, `GlStateManager.BLEND.mode.enabled` might be false even blend is enabled due to `GL32C.glEnablei()` call in `enableBufferBlend()`, and then GlStateManager will considered blend is already false, and finish `GlStateManager.BooleanState.setEnabled()` without calling `GL11.glDisable`.
`disableBufferBlend()` in same class have the same issue, tho it might not affect shaders so much like `enableBufferBlend()`, and disable single buffer blend might not need to set global blend to false.

IMS212:
Fixed; Iris will consider it's state "unknown" and honor any future set regardless of state.

PR #2114, opened 2023-08-31
Fix incorrect blend mode in post processing

moyongxin:
An ugly (and maybe improper) fix for #2113


IMS212:
I don't see why this should fix anything. Also, would you mind switching the code to RenderSystem?

moyongxin:
> I don't see why this should fix anything. Also, would you mind switching the code to RenderSystem?

Because RenderSystem._disableBlend() may defer the operation and fix nothing.

IMS212:
Applied a more correct fix based on GeforceLegend's suggestion in the issue.

PR #2137, opened 2023-10-03
Fix typos

omahs:
Fix typos


IMS212:
trying to farm I see

omahs:
@IMS212 farm what exactly? It seems that you don't always respond like that to relatively minor PR, is it because of hacktoberfest?

IMS212:
yes

omahs:
@IMS212 Ok, good to know. I often fix docs, hope not to have too many reactions like this. Wishing you a nice day

PR #2211, opened 2024-01-07
Fix iris website url in LaunchWarn

douira:
it was pointing to the old .net website.


IMS212:
not an issue in 1.20.3

PR #2225, opened 2024-01-29
Update build-release.yml

EchoEllet:
Update checkout and setup-java to GitHub actions to the latest versions


NathanHuisman (changes requested):
You can upgrade mc-publish to version 3.3 as well, as 3.2 is deprecated [(CI fails)](https://github.com/IrisShaders/Iris/actions/runs/7703861453)

PR #2238, opened 2024-02-06
Update bulid*.yml

LY-Xiang:
actions/* v2->v4
Kir-Antipov/mc-publish v3.2->v3.3

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -10,14 +10,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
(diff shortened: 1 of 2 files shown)


IMS212:
Um... this doesn't actually build.

Kichura (commented):
The checkout step has no "name" statement, meaning it has no friendly name.

LY-Xiang:
L13 
- name: Checkout sources

Kichura (approved):
Despite `check-latest: true` being excluded below `java-version: 17` for auto-updating reasons - lgtm.

Issue #1365, opened 2022-04-12
Horse markings render over horse armour


Traben-0:
### What happened?

with iris-mc1.18.2-1.2.2-build.32 all horses render their markings over the top of their armour

I have reproduced this consistently with only iris and sodium installed and no resourcepacks used.
simply load a world, tame a horse and place diamond horse armour on it

I don't know if this is relevant but horse armour uses the getEntityCutoutNoCull renderLayer and the markings use getEntityTranslucent, in vanilla atleast.

### Screenshots

![2022-04-12_20 51 17](https://user-images.githubusercontent.com/8456487/162946644-5d59d789-2b99-45ac-a653-23c43ae1b977.png)

### Minecraft Version

Minecraft 1.18.2

### Iris Version

iris-mc1.18.2-1.2.2-build.32.jar

### Sodium Version

sodium-fabric-mc1.18.2-0.4.1+build.15.jar

### Operating System

Windows 11

### What is your GPU?

Nvidia Geforce RTX 3080

### Java Version

Java 17

Traben-0:
also present in iris-mc22w15a-1.2.3-pre-f05b1251-dirty.jar

coderbot16:
> I don't know if this is relevant but horse armour uses the getEntityCutoutNoCull renderLayer and the markings use getEntityTranslucent, in vanilla atleast.

This is indeed relevant and the underlying issue is that Iris Batched Entity Rendering is trying to ensure that translucent things render after other things, in the hopes that it solves issues like translucent slime parts not blending correctly with other entities. 

However, in this case, it breaks since the translucent horse markings are being rendered after the non-translucent horse armor rather than the other way around.

Here's the part of the code that does the sorting based on the "transparency type":

https://github.com/IrisShaders/Iris/blob/fb22e63d775a8eaef9b184f1c6a04ddf5fa142a0/src/main/java/net/coderbot/batchedentityrendering/impl/ordering/GraphTranslucencyRenderOrderManager.java

O-d-i-n-n:
Same happens on the latest version (1.5.1) in 1.19.3

razorchase49:
Confirmed for version 1.5.2 for 1.19.4.
So, how does this get fixed? Is there some sort of patch, or do we need to wait for an update?

Lolothepro:
wait

razorchase49:
Confirmed for version 1.6.0

razorchase49:
Confirmed for version 1.6.1.
[...]

PR #2287, opened 2024-04-02
 Fix horse armor/markings render layers ordering

doonguy:
This fixes issue #1365 where horse markings would appear above horse armor. I'm quite the amateur so I don't know that this is the best fix, it's probably inefficient but it does seem to get the job done. Tested in 1.18.2 and 1.20.4, along with in my 1.20.4 modpack and I couldn't find any newly created issues.


doonguy:
Going to try to do what you suggested but base things off the refractor branch as I see that's the new default and I'm guessing easier to merge into

PR #2289, opened 2024-04-03
Fix horse armor/markings render layers ordering (refractor)

DoonGuy:
Fixes rendering order issues with entities such as horses by treating all objects drawn on an entity after a translucent object as also being translucent. Should be a cleaner implementation of #2287 using the method suggested there instead. I've tested it in the dev client with IntelliJ and everything seems to work properly. I wasn't able to get a working .jar, but I think that was related to the new build environment.

### DIFF
--- a/src/main/java/net/irisshaders/batchedentityrendering/impl/ordering/GraphTranslucencyRenderOrderManager.java
+++ b/src/main/java/net/irisshaders/batchedentityrendering/impl/ordering/GraphTranslucencyRenderOrderManager.java
@@ -20,7 +20,9 @@ public class GraphTranslucencyRenderOrderManager implements RenderOrderManager {
 	private final FeedbackArcSetProvider feedbackArcSetProvider;
 	private final EnumMap<TransparencyType, Digraph<RenderType>> types;
 	private final EnumMap<TransparencyType, RenderType> currentTypes;
+	private final List<RenderType> processedRenderTypes = new ArrayList<>();
 	private boolean inGroup = false;
+	private boolean translucentRendered = false;
 
 	public GraphTranslucencyRenderOrderManager() {
 		feedbackArcSetProvider = new SimpleFeedbackArcSetProvider();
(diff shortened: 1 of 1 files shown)


src/main/java/net/irisshaders/batchedentityrendering/impl/ordering/GraphTranslucencyRenderOrderManager.java
```diff
@@ -46,10 +48,21 @@ private static TransparencyType getTransparencyType(RenderType type) {
 	}
 
 	public void begin(RenderType renderType) {
-		TransparencyType transparencyType = getTransparencyType(renderType);
+
+		if (processedRenderTypes.contains(renderType)) {
+			return;
+		}
+
+		TransparencyType transparencyType = translucentRendered ? TransparencyType.GENERAL_TRANSPARENT : getTransparencyType(renderType);
```
coderbot16, line 56: Would probably be beneficial to add a small comment explaining why this is necessary for future readers.

coderbot16 (commented):
Nice work! A few small suggestions but the underlying changes are good.

doonguy:
Thanks for the feedback! Been testing a little bit more and I've found that the game crashes in certain scenarios with "Buffer has already been released!" I will look into it, but thought I should mention that sooner than later so this doesn't get merged.

coderbot16:
No problem and thanks for the heads up! It might be a problem if the single buffer for a given render type is getting drawn multiple times and the other parts of the code don't expect that. Not sure of the best solution off the top of my head.

DoonGuy:
I'm going to close this for now, but I do plan on learning some more and coming back to this. For now, I've made a separate mod [here](https://github.com/DoonGuy/Bad-Horse-Fix) that that fixes the issue the original way. That way people can take the efficiency hit only if they choose to since it's such a small thing that it fixes.

coderbot16:
Take as much time as you need! I don't think your old method really has a huge performance hit, but rather it's just that the bug probably affects more than just horses so a general solution would fix everything all at once. Regardless, it seems like you've got a pretty solid footing already with Minecraft/Java - excited to see what you figure out!

PR #2324, opened 2024-05-14
add boolean expression parser to disable programs

### LINKED ISSUE #1026: shaders.properties program toggles are not supported
These are declarations of the form:

```
program.world0/deferred1.enabled = HBAO || RSM
```

fayer3:
fixes #1026

adds a boolean expression parser. it might not be the best implementation, but it works from my testing.

the parser uses 2 Stacks, one for operations andone for the parsed values.
expressinos are read front to back, and evaluated back to front.
brackets and && chains are evaluated once they close/end

### DIFF
--- a/src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
+++ b/src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
@@ -1,0 +1,150 @@
+package net.irisshaders.iris.shaderpack.parsing;
+
+import net.irisshaders.iris.Iris;
+import net.irisshaders.iris.shaderpack.option.values.OptionValues;
+
+import java.util.EmptyStackException;
+import java.util.Stack;
+
+public class BooleanParser {
+	private enum OPERATION {
+		AND {
+			@Override
+			boolean compute(boolean value, Stack<Boolean> valueStack) {
+				return value & valueStack.pop();
+			}
+		}, OR {
+			@Override
+			boolean compute(boolean value, Stack<Boolean> valueStack) {
+				return value | valueStack.pop();
+			}
+		}, NOT {
+			@Override
+			boolean compute(boolean value, Stack<Boolean> valueStack) {
+				return !value;
+			}
+		}, OPEN;
+
+		boolean compute(boolean value, Stack<Boolean> valueStack) {
+			return value;
+		}
+	}
+
+	/**
+	 * parses the given expression
+	 * @param expression expression to parse
+	 * @param valueLookup lookup of shadow options
+	 * @return result of the expression, or true if there was an error
+	 */
+	public static boolean parse(String expression, OptionValues valueLookup) {
+		try {
+			String option = "";
+			Stack<OPERATION> operationStack = new Stack<>();
+			Stack<Boolean> valueStack = new Stack<>();
+			for (int i = 0; i < expression.length(); i++) {
+				char c = expression.charAt(i);
+				switch (c) {
+					case '!' -> operationStack.push(OPERATION.NOT);
+					case '&' -> {
+						// add value first, because this checks for preceding NOTs
+						if (!option.isEmpty()) {
+							valueStack.push(processValue(option, valueLookup, operationStack));
+							option = "";
+						}
+						// AND operators have priority, so add a bracket if it's the first AND
+						if (operationStack.isEmpty() || !operationStack.peek().equals(OPERATION.AND)) {
+							operationStack.push(OPERATION.OPEN);
+						}
+						i++;
+						operationStack.push(OPERATION.AND);
+					}
+					case '|' -> {
+						// add value first, because this checks for preceding NOTs
+						if (!option.isEmpty()) {
+							valueStack.push(processValue(option, valueLookup, operationStack));
+							option = "";
+						}
+						// if there was an AND before, that needs to be evaluated because it takes priority
+						if (!operationStack.isEmpty() && operationStack.peek().equals(OPERATION.AND)) {
+							evaluate(operationStack, valueStack, true);
+						}
+						i++;
+						operationStack.push(OPERATION.OR);
+					}
+					case '(' -> operationStack.push(OPERATION.OPEN);
+					case ')' -> {
+						// add value first, because this checks for preceding NOTs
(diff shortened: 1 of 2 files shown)


src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.util.EmptyStackException;
+import java.util.Stack;
+
+public class BooleanParser {
+	private enum OPERATION {
+		AND {
+			@Override
+			boolean compute(boolean value, Stack<Boolean> valueStack) {
+				return value & valueStack.pop();
```
douira: is there a reason for using bitwise instead of logical operators here?

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
```diff
@@ -0,0 +1,150 @@
+package net.irisshaders.iris.shaderpack.parsing;
+
+import net.irisshaders.iris.Iris;
+import net.irisshaders.iris.shaderpack.option.values.OptionValues;
+
+import java.util.EmptyStackException;
+import java.util.Stack;
+
+public class BooleanParser {
+	private enum OPERATION {
```
douira: enums in this codebase are typically in UpperCamelCase (though their elements are SCREAMING_SNAKE_CASE)

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
```diff
@@ -0,0 +1,150 @@
...
+							option = "";
+						}
+						// if there was an AND before, that needs to be evaluated because it added its own bracket
+						if (!operationStack.isEmpty() && operationStack.peek().equals(OPERATION.AND)) {
+							evaluate(operationStack, valueStack, true);
+						}
+						evaluate(operationStack, valueStack, true);
+					}
+					case ' ' -> {}
+					default -> option += c;
```
douira, line 88: using a StringBuilder instead of repeated string concatenation is probably more appropriate here.

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
```diff
@@ -0,0 +1,150 @@
...
+				}
+			}
+			if (!option.isEmpty()) {
+				valueStack.push(processValue(option, valueLookup, operationStack));
+			}
+			evaluate(operationStack, valueStack, false);
+			boolean result = valueStack.pop();
+			if (!valueStack.isEmpty() || !operationStack.isEmpty()) {
+				Iris.logger.warn(
```
douira, line 97: a LOGGER for this class could make these log messages more specific (it tags the messages with a class then). This is up to you

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	/**
+	 * gets the value for the given string and negates it if there is a NOT in the operationStack
+	 */
+	private static boolean processValue(String value, OptionValues valueLookup, Stack<OPERATION> operationStack) {
+		boolean booleanValue = switch (value) {
+			case "true", "1" -> true;
+			case "false" -> false;
```
douira: does a `0` literal also exist?

douira (changes requested):
the negation handling is a little funky and I have a few minor nitpicks but it looks good to me overall! Thanks for the work.

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
fayer3: logical && or || will shortcircuit and not pop the stack when the value is already true

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
fayer3: don't think so, the "1" is just there, because the preprocessor replaces stuff that is defined with a 1

fayer3: I can add a "0" aswell though

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
fayer3: maybe, I just thought since it's reset often I would just use Strings

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
douira: Ah right I remember that effect

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
fayer3: will change that

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
douira: yeah good point, probably doesn't end up making much of a difference here since both operations happen.

src/main/java/net/irisshaders/iris/shaderpack/parsing/BooleanParser.java
fayer3: I could reverse the order then it doesn't matter

IMS212:
There's a important question here; what's stopping JCPP from automatically doing all of this?

fayer3:
jcpp only replaces the options
afaik it only evaluates boolen expressions in #ifs

fayer3:
you could maybe do some regex to convert
```
program.X.enabled=A && B || C
```
to 
```
#if !(defined A && defined B || defined C)
program.X.enabled=false
#endif
```
that should then work with jcpp

IMS212:
Thanks!

PR #2336, opened 2024-06-05
add shader name to compatibility patcher log messages

fayer3:
adds a name to the patcher Parameters, to be able to log the name when compatibility patching

not sure if that is the best way to do it.
the transform calls now have the name twice, once from the name argument, and once from the paramters name


douira:
That is indeed not very elegant. Maybe the field could be set inside the `transform` method using the name already provided as a parameter to the `transform` method? This would also avoid plumbing the name through all the constructors.

Thanks for the work.

fayer3:
could do that yeah

fayer3:
changed it to be set in the TransformPatcher right before the transformer call

douira (approved):
looks good to me

PR #2350, opened 2024-06-13
fix const float parsing

fayer3:
fixes the const float parser failing for stuff with the `f` suffix, or no characters after the semicolon, like
```
const float ambientOcclusionLevel = 1.0f;   //[0.0f 0.1f 0.2f 0.3f 0.4f 0.5f 0.6f 0.7f 0.8f 0.9f 1.0f]
```
or
```
const float ambientOcclusionLevel = 0.5;
```


fayer3:
also fixes defines with float suffix like 
```
#define AO 1.0f   // [0.0f 0.1f 0.2f 0.3f 0.4f 0.5f 0.6f 0.7f 0.8f 0.9f 1.0f]
```

PR #2378, opened 2024-06-22
NullPointerException

Mathaeux:


fayer3:
what exactly is the issue with that? according to the docs this is allowed to be null

PR #2400, opened 2024-07-04
Update supported versions in readme

llakala:
The list of current versions was outdated and implied that unmaintained releases were still being maintained. I updated the current list, and added a statement that while previous versions do exist, support won't be provided for them.


llakala:
Bumping this :3

PR #2421, opened 2024-07-20
Implement new IRIS_VERSION gl define

ArkoSammy12:
This PR adds a new IRIS_VERSION gl define. It gets the Iris version using `Iris#getVersion`, throwing if it returns `null`, then attempts to match the semver using the `SEMVER_PATTERN`, throwing if it can't be matched. The logic to turn a semver into the 122 format has been extracted into its own method, and `StandardMacros#getMcVersion` has been modified accordingly.

Feel free to suggest or make any changes as necessary.


IMS212:
Looks good to me!

PR #2457, opened 2024-08-27
fix capitalization check for zipped shaders

fayer3:
adds a `isZip` parameter to ShaderPack and `IncludeGraph` to skip the file capitalization check, since it cause an exception for zips


fayer3:
fixed an issues when using symbolic links that redirect anything before the `shaders` folder

PR #2483, opened 2024-10-04
add missing internal texture formats and pixel types

fayer3:
didn't have a setup to check if they actually work
not sure if the RGBA2 is even useable


Luracasmus:
Why are the checks failing?

fayer3:
because the iris build system expects files that are not in the repo

PR #2592, opened 2025-01-05
Fix hard coded vertex format element IDs

### LINKED ISSUE #2491: Static initialization of VertexFormatElements is incompatible with other mods
### What happened?

The game crashes when multiple mods try to add new `VertexFormatElement`s if their IDs overlap with Iris'.
Additionally, due to the parallel nature of Mod Startup in NeoForge it probably isn't threadsafe.

The advice I got on what to pass to the ID parameter was to AT the `ELEMENTS` array in `VertexFormatElement` and use its `size()` method; doing so, on the NeoForge side, in `FMLClientSetupEvent#enqueueWork`, so it's thread safe.

Additionally XFactHD mentioned that the passed values to the `index` parameter are not correct

https://github.com/IrisShaders/Iris/blob/d6bf750fe256c6be54c560de543892f1f00e9654/common/src/main/java/net/irisshaders/iris/vertices/IrisVertexFormats.java#L20-L24

Apparently, what should be passed there is how often the specific USAGE has occurred before, so you'd have to count through the `ELEMENTS` array how many are there already.

Here's the direct quote from the Discord:
> What should be passed to the `index` parameter?

> The occurence of the "element usage". It's 0 for most things, but UV for example has indices 0, 1 and 2 for texture, hurt overlay and light respectively. They'd have to count how many elements with any given usage are already registered and use that as the index (you should ideally do that too)

### Screenshots

_No response_

### Log output

https://0x0.st/X69G.txt

### Minecraft Version

1.21.1

### Iris Version

iris-neoforge-1.8.0-beta.5

### Operating System

Arch Linux

### What is your GPU?

AMD Radeon 7900XT

### Additional context

See [discussion on the NeoForge discord](https://discord.com/channels/313125603924639766/1249305774987939900/1294708936439631913)

I discovered this while trying [my port to 1.21.1 of Stellar View](https://github.com/Povstalec/StellarView/pull/38) with Iris

Erdragh:
This PR changes the registration of the `VertexFormatElement`s to not use hardcoded IDs and instead tries to fit the IDs into gaps of the ID -> VFE mapping.

This prevents crashes on startup with other mods that handle new `VertexFormatElement`s the same way (i.e. Stellar View), which would previously try to register with IDs that were already taken by Iris.

Fixes #2491

### DIFF
--- a/common/src/main/java/net/irisshaders/iris/vertices/IrisVertexFormats.java
+++ b/common/src/main/java/net/irisshaders/iris/vertices/IrisVertexFormats.java
@@ -26,11 +26,11 @@ public class IrisVertexFormats {
 			}
 		}
 
-		ENTITY_ELEMENT = VertexFormatElement.register(10, 0, VertexFormatElement.Type.SHORT, VertexFormatElement.Usage.GENERIC, 2);
-		ENTITY_ID_ELEMENT = VertexFormatElement.register(11, LAST_UV + 1, VertexFormatElement.Type.USHORT, VertexFormatElement.Usage.UV, 3);
-		MID_TEXTURE_ELEMENT = VertexFormatElement.register(12, 0, VertexFormatElement.Type.FLOAT, VertexFormatElement.Usage.GENERIC, 2);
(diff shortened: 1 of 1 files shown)


IMS212:
Looks good to me at first glance.

PR #2629, opened 2025-02-13
Fixed Tri.equals()

Luna5ama:
This should fix the transform cache


IMS212:
doesn't this die if the first one has null elements (Object.equals should solve that)

Luna5ama:
> doesn't this die if the first one has null elements (Object.equals should solve that)

You're so right, fixed.

PR #2630, opened 2025-02-13
Parallel preprocessor processing

Luna5ama:
It makes packs load faster (depending on how many preprocessors they spammed)

### DIFF
--- a/common/src/main/java/net/irisshaders/iris/shaderpack/ShaderPack.java
+++ b/common/src/main/java/net/irisshaders/iris/shaderpack/ShaderPack.java
@@ -67,6 +67,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


douira:
It would be interesting to actually see a profile of shader pack loading before and after this patch. My past measurements (quite some time ago, so they might now be outdated) indicate the preprocessor's share of the total shader reloading time is very small.

Luna5ama:
> It would be interesting to actually see a profile of shader pack loading before and after this patch. My past measurements (quite some time ago, so they might now be outdated) indicate the preprocessor's share of the total shader reloading time is very small.

Like the description said. It depends on how many preprocessors they use. I saw good gains on Photon and [Alpha Piscium](https://github.com/Luna5ama/Alpha-Piscium/pulse). Also, there is #2629 that fixes transform cache.

SpacEagle17:
Any updates on this? Would be interesting to see if it could be looked into if the claims are correct :)

Luna5ama:
Ok so right now I am having loading performance issues with my pack then I remembered this exist. So on my pack's latest commit ([d2d699ba](https://github.com/Luna5ama/Alpha-Piscium/commit/d2d699ba8c7ae11a6f84e049e14f4121c5427acc)). Preprocessing take a long time. This did make it faster.

I reload my pack 15 times while running a profiler and this is the result:
Before:
image
image
After:
image
image

As you can see from the second screenshot of each, there is some parallelism going on so it is working.
Reload time:
Before: ~750ms
After: ~250ms

PR #2728, opened 2025-04-26
Mexican Spanish Translation

TheLegendofSaram:
Just a small translation for the mod, should work for the other spanish variants


TH3N3WN00B:
Español speaking here, lgtm

PR #2840, opened 2025-08-19
Fixed blending behaviour for Shadowcomp passes

rre36:
Shadowcomp passes got force-enabled blending sometime after 1.21.5's Iris release for... reasons(???).
Did a simple "fix" by matching the behaviour of the blending setup of regular composite passes. Not sure if there's a better way to do this but it seems to work and might also come with the added benefit of supporting blending controls by the shaderpack for shadow composites (did not verify this because I dont need it).


rre36:
Possible underlying cause might be Blaze3D's withoutBlend() call not working as expected, which is being masked by the additional blending setup done on regular composite passes.

rre36:
Addendum: Shadowcomp passes do indeed support blending controls via shaders.properties with this change.

IMS212:
Looks good to me. I believe I know why the withoutBlend call doesn't work; it's a change done by Iris further down. But this fixes it.

Issue #2798, opened 2025-06-29
at_tangent.w is sometimes -1.0 when it should be 1.0


Kirilpro6284:
### What happened?

(look at top left corner)

top value = at_tangent.w
bottom value = dot product between vertex normal and cross(bitangent, tangent) (both derived from dFdx/y(texcoord))

doesn`t happen in optifine

### Screenshots

![Image](https://github.com/user-attachments/assets/fb545f42-a1a2-496a-8731-3219d7e6b04d)
![Image](https://github.com/user-attachments/assets/b470857c-ea9e-40ec-ba49-c34ee9a76613)

### Log output

[latest.log](https://github.com/user-attachments/files/20970224/latest.log)

### Minecraft Version

1.21.1

### Iris Version

iris-fabric-1.8.8+mc1.21.1

### Operating System

windows 10

### What is your GPU?

RTX 4060

### Additional context

```glsl
mat2x3 pixelsToPlayerPos = mat2x3(dFdx(eyePlayerPos), dFdy(eyePlayerPos));
mat2 pixelsToTexCoord = mat2(dFdx(texcoord), dFdy(texcoord));
mat2 texCoordToPixels = inverse(pixelsToTexCoord);

vec3 tangent         = normalize(pixelsToPlayerPos * texCoordToPixels[0]);
vec3 bitangent       = normalize(pixelsToPlayerPos * texCoordToPixels[1]);

float handedness = sign(dot(normal, cross(bitangent, tangent)));
```
(this is how i recalculate the correct handedness)

Kirilpro6284:
![Image](https://github.com/user-attachments/assets/895b8034-7a5f-47d6-89d0-2885c43958cd)
Here's an example of broken normal mapping on some blocks as a result of this.

Issue #2709, opened 2025-04-14
Water top normals are off by ~0.455°


Slaynash:
### What happened?

I was investigating a reflection issue on Complementary earlier, and discovered that the value of `gl_Normal` in gbuffers_water.vsh was slightly off by ~0.455° (z component < 0). I'm not sure if this is expected or not but it's worth noting.

### Minecraft Version

1.21.4

### Iris Version

1.8.8

### Operating System

Windows 11

### What is your GPU?

RTX 4070

### Additional context

- Sodium 0.6.13
- Distant Horizons 2.3.2-b

GeForceLegend:
Duplicate of #2554

Issue #2554, opened 2024-12-04
Incorrect vertex normal for terrain


GeForceLegend:
### What happened?

Iris is using unsigned normalized octahedron decoding for vertex normal. To get exact vertex normal for common surfaces, it needs the input be 0.0 and 0.5. But in vertex attributes, Iris is using RGBA8_SNORM and even throw values between [-1.0, 0.0) away, and 0.5 will be about 0.496 in vertex attribute. This will create obvious artifact in some situation, like this image using ComplementaryReimagined_r5.2.2, because of the incorrect normal, left part is reflecting things under water:

![Image](https://github.com/user-attachments/assets/29769842-63e7-449c-b5d6-ad7dce8920f8)

As Iris is using RGBA8_SNORM as vertex attribute format, why not just using the full [-1.0, 1.0] range, and removing `* 2.0 + 1.0` in the beginning of octahedron decoding code? RGBA8_SNORM can get 0.0 correctly (works same as 0.5 in current solution), which means we will get exact normal data on common surfaces.

### Minecraft Version

Minecraft 1.21.3

### Iris Version

iris-fabric-1.8.1+mc1.21.3.jar

### Operating System

Windows 10

### What is your GPU?

NVIDIA GeForce RTX 4070

GeForceLegend:
Confirmed to be fixed in iris-fabric-1.9.3+mc1.21.9

PR #2847, opened 2025-08-26
Fixed normal and tangent vectors encoding

lipesto:
This change adresses following issues:
https://github.com/IrisShaders/Iris/issues/2798
https://github.com/IrisShaders/Iris/issues/2709
https://github.com/IrisShaders/Iris/issues/2554

This change 
Removes incorrect edge case of handedness encoding.
Doubles precision of normal and tangent attributes.
Adds explicit zero and one values, allowing stored vectors to be aligned with world grid.


IMS212:
Thanks!

PR #2853, opened 2025-09-04
Use appropriate shader for custom dimension types with nether or end effects

### LINKED ISSUE #2200: End shader no longer working for custom dimension types
### What happened?

I have a world on my server that uses a custom dimension type.
The dimension type is added via a datapack and is configured so that is has the end sky enabled:
```jsonc
{
  "effects": "minecraft:the_end",
  // more properties ...
}
```

The end dimension shader was enabled by Iris just fine.
Recently though, I upgraded Iris from 1.6.4 to 1.6.11.
Since then, the overworld shader is enabled instead.

### Screenshots

Before the upgrade (Iris 1.6.4): 
iris 1.6.4

After the upgrade (Iris 1.6.11):
iris 1.6.11

### Log output

[latest.log](https://github.com/IrisShaders/Iris/files/13759924/latest.log)

### Minecraft Version

Minecraft 1.20.1

### Iris Version

iris-mc1.20.1-1.6.11.jar

### Operating System

Windows 10

### What is your GPU?

Nvidia GeForce RTX 4070

### Additional context

I modified the NBT data of the world I am using. This way the overworld of my server lobby can have a different dimension type, like so:
image

Upon a little search, I found [this commit](https://github.com/IrisShaders/Iris/commit/26a323686958bebec59bbcd128defa2717f64759#diff-81965af5d12f85cd5aeac5544a827d132d205e26d38e098b3c16013b8a2ef610L640-L646) which is likely the reason.
It seems the hardcoded check for the end and nether effects were removed and replaced with a shader-specific `dimension.properties` file.

I tried modifying the file, but it seems the lookup is done only for the _dimension_ of the world.
Is there a way to configure _dimension types_ instead?

Also, it seems very impractical to me, that I have to patch shader packs in order to get the proper shader for my world...
Is there a better way?

LCLPYT:
Closes #2200 

If you enable an Iris shader in a dimension with a custom dimension type using the nether or end effects, you get a buggy result:
buggy_result
The overworld shader is used but the dimension effects are also rendered at the same time, in this case the end sky.
As described in https://github.com/IrisShaders/Iris/issues/2200, this was broken by [this commit](https://github.com/IrisShaders/Iris/commit/26a323686958bebec59bbcd128defa2717f64759#diff-81965af5d12f85cd5aeac5544a827d132d205e26d38e098b3c16013b8a2ef610L640-L646) and was working fine in Iris 1.6.4 and before.

This is how it looks in vanilla (just an empty custom dimension with the end sky):
no_shader

This PR reimplements the check for the dimension type effects in `Iris::getCurrentDimension`.
If the end effect is active, use the end dimension shader by assuming we are in the end dimension.
If the nether effect is active, assume we are in the nether dimension.

With the check the result looks as expected:
fixed

Could  there be any problems with overriding the current dimension id this way?
IMO most of the time the nether / end effect is desired when having a custom dimension with that sky effect.

The fix can be tested by using this datapack, for example:
[iris_bug_demo.zip](https://github.com/user-attachments/files/22139156/iris_bug_demo.zip)

Just install it when creating a new world and teleport to the dimension with this command:
```
/execute in minecraft:iris_bug_demo run tp @p 0 100 0
```

### DIFF
--- a/common/src/main/java/net/irisshaders/iris/Iris.java
+++ b/common/src/main/java/net/irisshaders/iris/Iris.java
@@ -45,6 +45,8 @@ import net.minecraft.client.multiplayer.ClientLevel;
 import net.minecraft.network.chat.ClickEvent;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.HoverEvent;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.level.Level;
 import org.jetbrains.annotations.NotNull;
 import org.lwjgl.glfw.GLFW;
 import org.lwjgl.opengl.ARBParallelShaderCompile;
@@ -611,6 +613,21 @@ public class Iris {
 		ClientLevel level = Minecraft.getInstance().level;
 
 		if (level != null) {
+			// Check if the dimension type of the current level has custom effects set (end sky or nether).
+			// This is minecraft:overworld by default, but can also be minecraft:the_nether or minecraft:the_end.
+			// The appropriate shader for the dimension should be used by default in order to prevent buggy results.
+			// More information at https://minecraft.wiki/w/Dimension_type
+			// https://github.com/IrisShaders/Iris/issues/2200
+			ResourceLocation effects = level.dimensionType().effectsLocation();
+
+			if (Level.END.location().equals(effects)) {
+				return DimensionId.END;
+			}
+
+			if (Level.NETHER.location().equals(effects)) {
+				return DimensionId.NETHER;
+			}
+
 			return new NamespacedId(level.dimension().location().getNamespace(), level.dimension().location().getPath());
 		} else {
 			// This prevents us from reloading the shaderpack unless we need to. Otherwise, if the player is in the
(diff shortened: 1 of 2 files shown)


SpacEagle17:
This seems to be a clever approach to handle this. Ideally though there should be a check if the current dimension doesn't already have a dedicated handling in dimension.properties.

LCLPYT:
Thanks for your feedback, I agree. I've added a check if there is an **exact** match in dimension.properties.
I tested it with Complementary Unbound, which has the following entries:
image
As you can see, if wildcards were to be matched, the overworld shader would apply to every custom dimension, which just leads to the current situation again... This is why only exact matches are used.
Do you have a better idea maybe?

SpacEagle17:
I think an exact match makes sense. Imagine a shader made a `dimension.void = the_deeper_void:void_dimension` entry in the dimensions.properties file and let's just assume it uses the end effect but the shader already has proper handling for it which explains said above created entry. We basically don't want to touch it then and not override it with what you did. That was my thought process .

Basically only handle unknown dimensions, ones which aren't in the properties file via the your dimension effect fallback

IMS212:
Thanks! I've merged this myself into the default branch.

PR #2856, opened 2025-09-08
Allow sampling of depthtex2 in gbuffers

jbritain:
When rendering the hand, the developer may wish to sample the depth behind the hand.


IMS212:
i feel like this is a bad idea, but fine

PR #2857, opened 2025-09-09
Removed unnecessary full directory tree checks for shader directory for non-zip shader pack

Luna5ama:
The actual pack loading for non-zip shader pack doesn't look into full directory tree anyways so why do it. It is slower to go through the whole directory tree especially when there are tons of files/directory, such as stuff in .git.


SpacEagle17:
Ngl this should be merged

PR #2955, opened 2025-12-23
Fix custom dimension skybox (1.21.11 regression)

### LINKED ISSUE #2200: End shader no longer working for custom dimension types

LCLPYT:
The end shader is no longer used for custom dimensions with the end skybox.

image

This was caused because of a change in the 1.21.11 upgrade in commit 7f17ac47e027c72326678ae31549c9cc6de9493d in Iris::getCurrentDimension:

```diff
-			ResourceLocation effects = level.dimensionType().effectsLocation();
+			Identifier effects = level.dimension().identifier();
```
This broke the previous patch from PR #2853 

In 1.21.11, the `effectsLocation()` property is now called `skybox()` and now returns a Skybox enum constant. Possible values are: `NONE`, `OVERWORLD` and `END`

This PR updates the dimension overwrite for the `END` and `OVERWORLD` skyboxes. As there is no nether skybox option, the nether overwrite was removed.

Closes #2200 

Updated datapack for testing: [iris_bug_demo.zip](https://github.com/user-attachments/files/24318265/iris_bug_demo.zip)

image

### DIFF
--- a/common/src/main/java/net/irisshaders/iris/Iris.java
+++ b/common/src/main/java/net/irisshaders/iris/Iris.java
@@ -46,7 +46,7 @@ import net.minecraft.network.chat.ClickEvent;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.HoverEvent;
 import net.minecraft.resources.Identifier;
-import net.minecraft.world.level.Level;
+import net.minecraft.world.level.dimension.DimensionType;
 import org.jetbrains.annotations.NotNull;
 import org.lwjgl.glfw.GLFW;
 import org.lwjgl.opengl.ARBParallelShaderCompile;
@@ -623,19 +623,19 @@ public class Iris {
 				return dimensionId;
 			}
 
-			// Check if the dimension type of the current level has custom effects set (end sky or nether).
-			// This is minecraft:overworld by default, but can also be minecraft:the_nether or minecraft:the_end.
+			// Check if the dimension type of the current level has a custom skybox set (end sky or overworld).
+			// This is OVERWORLD by default, but can also be END or NONE.
 			// The appropriate shader for the dimension should be used by default in order to prevent buggy results.
 			// More information at https://minecraft.wiki/w/Dimension_type
 			// https://github.com/IrisShaders/Iris/issues/2200
-			Identifier effects = level.dimension().identifier();
+			DimensionType.Skybox skybox = level.dimensionType().skybox();
 
-			if (Level.END.identifier().equals(effects)) {
+			if (skybox == DimensionType.Skybox.END) {
 				return DimensionId.END;
 			}
 
-			if (Level.NETHER.identifier().equals(effects)) {
-				return DimensionId.NETHER;
+			if (skybox == DimensionType.Skybox.OVERWORLD) {
+				return DimensionId.OVERWORLD;
 			}
 
 			return dimensionId;


TehBrian:
Could this end up somehow being fixed for the nether as well? I'm on 26.2 and experiencing that popular shaders (Bliss, BSL, Complementary) are using overworld effects for my custom nether dimension type.

LCLPYT:
I have the same problem with one of my dimensions.
Afaik, there is no simple marker that a dimension is using the same settings as the nether anymore.
In the past, there were these "effects" but those were removed in favor of individual aspects such as the skybox etc.

This fix currently only looks at the value for the skybox, as that is as close as a marker as I could find.
Sadly, there is no skybox for the nether, only one for the overworld and one for the nether...

I guess you could somehow check if the dimension settings are similar to those of the nether and use that as an indicator. However its not really that straightforward anymore I think...

TehBrian:
Maybe it'd be a good idea to recommend to shaderpack authors to check for `*:*nether*` (though I'm not sure that syntax is supported) in their `dimension.properties` rather than specifically `minecraft:the_nether`.

Would it be a bad idea to assume that the `none` skybox correlates to the nether? That's true for vanilla, at least.

This may be dumb, but in dimension types, `cardinal_light` has options `default` and `nether`. Could it be safe to assume that `nether` indicates that the world is intended to be nether-esque?

If only there were some way to communicate arbitrary data in dimension types. Maybe the name (`*:*nether*`) is as close as we can get.

TehBrian:
> Would it be a bad idea to assume that the `none` skybox correlates to the nether? That's true for vanilla, at least.
> 
> This may be dumb, but in dimension types, `cardinal_light` has options `default` and `nether`. Could it be safe to assume that `nether` indicates that the world is intended to be nether-esque?

Implemented in https://github.com/IrisShaders/Iris/pull/3276 :D

PR #3027, opened 2026-02-22
fix cloudTime uniform

fayer3:
1.21.11 changed the cloud time to be based on the levels gameTime instead of overall game ticks.
this changes the `cloudTime` uniform to reflect that new change


Samalando:
AH WHOOPS

Issue #3018, opened 2026-02-12
Flip issue with colortex containing raw textures


GeForceLegend:
### What happened?

In [this commit](https://github.com/GeForceLegend/Sundial-Lite/commit/c9eb6160d431c74a3777ee3feec97b35a3b4652e), I added atmosphere trasmittance LUT for Sundial Lite, using some flip to get it available through colortex7 while avoid affecting TAA after. This works well on Optifine, however, after the manually flip, Iris still feed colortex7 with that LUT, causing TAA completely broken.

This is what I do:
1. `flip.composite_pre.colortex7 = true`
2. `texture.composite.colortex7 = textures/AtmosphereTransmittance.bin TEXTURE_2D RGB16 256 256 RGB UNSIGNED_SHORT`
3. `flip.composite5.colortex7 = true`

We expect, and in Optifine it is, composite - compisite5 will get the transmittance LUT through colortex7, and after composite6, colortex7 will turns normal (TAA color and previous water depth). But in Iris, as described above, we still get the LUT from colortex7 even after composite6.

This issue is exclusive on raw texture. Removing attach colortex7 with the LUT, or even just replace the LUT with `texture.composite.colortex7 = minecraft:textures/atlas/blocks.png`, TAA will still working right.

### Minecraft Version

Minecraft 1.21.11

### Iris Version

iris-fabric-1.10.5+mc1.21.11.jar

### Operating System

Windows 10

### What is your GPU?

RTX 4070

PR #3040, opened 2026-03-10
fix custom texture handling in ShaderProperties

OrzMiku:
Remove unnecessary texture patching indirection by directly storing raw texture definitions in customTextures map instead of creating intermediate customtex samplers. This commit will fix https://github.com/IrisShaders/Iris/issues/3018

### DIFF
--- a/common/src/main/java/net/irisshaders/iris/shaderpack/properties/ShaderProperties.java
+++ b/common/src/main/java/net/irisshaders/iris/shaderpack/properties/ShaderProperties.java
@@ -468,7 +468,14 @@ public class ShaderProperties {
 					} else if (parts.length == 7) {
 						// 2D texture handling
 						type = TextureType.valueOf(parts[1].toUpperCase(Locale.ROOT));
-						irisCustomTextures.put(newSamplerName, new TextureDefinition.RawDefinition(parts[0], TextureType.valueOf(parts[1].toUpperCase(Locale.ROOT)), InternalTextureFormat.fromString(parts[2]).orElseThrow(IllegalArgumentException::new), Integer.parseInt(parts[3]), Integer.parseInt(parts[4]), 0, PixelFormat.fromString(parts[5]).orElseThrow(IllegalArgumentException::new), PixelType.fromString(parts[6]).orElseThrow(IllegalArgumentException::new)));
+						TextureDefinition definition = new TextureDefinition.RawDefinition(parts[0], TextureType.valueOf(parts[1].toUpperCase(Locale.ROOT)), InternalTextureFormat.fromString(parts[2]).orElseThrow(IllegalArgumentException::new), Integer.parseInt(parts[3]), Integer.parseInt(parts[4]), 0, PixelFormat.fromString(parts[5]).orElseThrow(IllegalArgumentException::new), PixelType.fromString(parts[6]).orElseThrow(IllegalArgumentException::new));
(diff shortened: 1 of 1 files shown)


OrzMiku:
If I understand correctly, using `irisCustomTextures` storage applies to all stages, and the old code achieves the effect of "only applying to specific stages" by replacing `<buffername>` with `customtexN` in the corresponding `<stage>`.

However, this leads to issues like https://github.com/IrisShaders/Iris/issues/3018, where `flip.<program>.<buffername>` would directly stop working because `<buffername>` has been replaced with `customtexN`.

Now, the original textures declared via `texture.X.Y = ...` will be placed directly into `customTextures[stage][samplerName]` — handled exactly the same way as PNG textures. No composite names are generated, no `customTexturePatching` is performed, and shader source code is no longer renamed.

Since `ShaderPack.readTexture()` can already handle all texture types (1D, 2D, 3D, rectangle) of `TextureDefinition.RawDefinition`, no downstream code modifications are required. With the original textures now registered under their real sampler names in stage-specific maps, the existing flip mechanism will work properly.

xkzs2007 (approved):

GeForceLegend:
This will break shader packs using suffixes binding multiple texture with different type to one colortex, for example:
```properties
texture.deferred.colortex8.0=texture/AtmoData_620_540_440_n4.bin TEXTURE_3D RGBA32F 512 128 33 RGBA FLOAT
texture.deferred.colortex8.1=texture/itt320.png
```

This is a feature in Optifine document https://github.com/sp614x/optifine/blob/master/OptiFineDoc/doc/shaders.properties#L127-L133

Considering Iris handles this properly before, I'll guess Iris can correctly detect the sampler type used in shaders, and patch them to different `customtex`. And then, we should get expected result by just apply similar edit in this pr only when texture type is `TEXTURE_2D`

OrzMiku:
When the original texture is 3D/1D, sampler3D/1D xxx will be renamed to customtexN, which does not affect the original sampler2D, so flipping the original sampler2D is fine.
Therefore, only 2d texture needs to be modified to use customTextures.

I also tested the edge cases mentioned by [GeForceLegend](https://github.com/GeForceLegend), and it seems to be fine.

GeForceLegend:
TEXTURE_RECTANGLE type shares the same length of TEXTURE_2D. That's why I creates an branch to detect if type is 2D.

OrzMiku:
> TEXTURE_RECTANGLE type shares the same length of TEXTURE_2D. That's why I creates an branch to detect if type is 2D.

---

If TEXTURE_RECTANGLE is also placed into customTextures, it becomes impossible to handle situations where both TEXTURE_2D and TEXTURE_RECTANGLE are present.

```
texture.deferred.colortex8.0=xxx TEXTURE_2D ...
texture.deferred.colortex8.1=yyy TEXTURE_RECTANGLE ...
```

Only one of them will take effect.

---

If only TEXTURE_2D is placed into customTextures, while TEXTURE_RECTANGLE remains in the original irisCustomTextures + patch, then if TEXTURE_RECTANGLE exists, the flip issue will still occur.

---

I have another branch here that completely abandons irisCustomTextures+customTexturePatching and uses other methods to work correctly: https://github.com/OrzMiku/Iris/tree/test
However, there are quite a few changes, and since I am not very familiar with Iris, the modifications are a bit messy.

PR #3052, opened 2026-03-26
Fix inadequate shadow matrices for a Zero-to-One backend

EnnuiL:
A straightforward fix that partially fixes Longview's zero-to-one depth code (Technique 1)

This has been tested with a copy of Longview that had all the Reverse Z code disabled since that has a separate set of unique issues

<details>
<summary>Comparison ([0, 1])</summary>

Before:
Longview Technique 1 pre-fix

After:
Longview Technique 1 post-fix

</details>

<details>
<summary>Reference ([-1, 1])</summary>

No Longview, either with or without fix

</details>

### DIFF
--- a/common/src/main/java/net/irisshaders/iris/shadows/ShadowMatrices.java
+++ b/common/src/main/java/net/irisshaders/iris/shadows/ShadowMatrices.java
@@ -1,6 +1,7 @@
 package net.irisshaders.iris.shadows;
 
 import com.mojang.blaze3d.platform.InputConstants;
+import com.mojang.blaze3d.systems.RenderSystem;
 import com.mojang.blaze3d.vertex.PoseStack;
 import com.mojang.math.Axis;
 import net.irisshaders.iris.Iris;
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/irisshaders/iris/shadows/ShadowMatrices.java
```diff
@@ -19,21 +20,22 @@ public class ShadowMatrices {
 
 	public static Matrix4f createOrthoMatrix(float halfPlaneLength, float nearPlane, float farPlane) {
 		//System.out.println("making a matrix with " + nearPlane + " / " + farPlane + " * " + halfPlaneLength);
-		return new Matrix4f().setOrthoSymmetric(halfPlaneLength * 2, halfPlaneLength * 2, nearPlane, farPlane);
+		return new Matrix4f().setOrtho(halfPlaneLength, halfPlaneLength, halfPlaneLength, halfPlaneLength, nearPlane, farPlane, RenderSystem.getDevice().isZZeroToOne());
```
EnnuiL: For clarity: yes, switching to `setOrtho` is necessary for the fix and if you switch back to `setOrthoSymmetric` you will break it!

Yay!

EnnuiL:
Nevermind, this needs further work.. this will break shadows on [-1, 1]

<details>

Before (Reference):
Screenshot_20260325_234028

After (Reference):
Screenshot_20260325_233952

</details>

EnnuiL:
After some discussion, I decided to keep this PR only to the fixing part itself, which doesn't do much, but fixes flickering on movement as an issue.

Further debugging will need RenderDoc, which uh, does not work on my machine anymore! (it's broken on linux)

PR #3062, opened 2026-03-30
Make render layers in block.properties working in 26.1

ZhuRuoLing:
Grab a `BlockState` from `AbstractBlockRenderContext` to fetch an render layer override in `WorldRenderingSettings.INSTANCE.getBlockTypeIds()`.
Tested with Sundial Alpha Build 2026-03-27.

Before:
image
After:
image


GeForceLegend:
This may better switching to `ModifyArg`? As IMS is busy on Sodium and Aperture, we may can't expect they improve this by their own after merging prs.

PR #3061, opened 2026-03-30
Fix #2487 (Light level comes with minimum 8 instead of 0)

GeForceLegend:
Fixed #2487 according to JellySquid3's suggestion in [Sodium#2761](https://github.com/CaffeineMC/sodium/issues/2761)

PR #3091, opened 2026-04-20
Fix lightmap clamp

Merlin1809:
The texture matrix already adds ``0.03125f`` to the lightmap coords so the lowest possible value was 0.06 instead of the expected 0.03

thanks to cortex for pointing that out


Merlin1809:
oh my god I missed that: https://github.com/IrisShaders/Iris/pull/3061/changes/0c5b130d70f9729dac20540bacbf511e816289fa

😭 
well, the other fix should still be good tho I guess...

PR #3106, opened 2026-05-05
Fixed selected shader not in menu center

GeForceLegend:
When creating shader selection menu, Iris will directly do the below thing when detecting the currently added one is selected pack:
```java
setSelected(entry);
setFocused(entry);
centerScrollOn(entry);
setApplied(entry);
```
However, this is done even before the current entry is added. So when selected pack detected, Iris will just roll current screen to the end, and then add the entry of selected pack. This will result in selected pack is perfectly outside the bottom of shaderpack selection menu when open it, which is very annoying.

By temporary store the matched entry, and do the above thing after all entry is added, now selected shader will be in the middle of menu when open it.


SpacEagle17:
Mind showing a screenshot of before and after? Couldn't visualize it from your description

PR #3168, opened 2026-06-21
Added FISM optimisations, untested due to Iris code not building

maDU59:
All needed informations are available here:

https://modrinth.com/mod/fism

https://github.com/maDU59/FasterIrisShadowMapper/issues/4


SpacEagle17:
Some packs might want these things rendered in the shadow pass though. 
If anything, a new shaders.properties argument has to be added to disable this functionality, so packs can get the previous behaviour if needed

And regarding not being able to build iris, what are the issues? Usually, I just uncomment Neoforge from the settings.gradle task and ensure the right sodium version is the dependency. That does it usually

maDU59:
I thought the same as you, that's why I originally made it an optional mod, but I talked about it with IMS before hand and here's what he said:
> I think it could be worth making a PR, looking at it now

> as long as it's just opaque objects, I don't think it needs a toggle

As for the building process, I just thought the repo was not in the build-ready process, I didn't know it was an easy fix. It should work anyway given how small the code is, at worst there's a missing import somewhere

maDU59:
I saw you merged that, nice!
Also I just noticed I forgot to update the injectors names, they are called `fism$foo` when it should probably be `iris$foo`

PR #3173, opened 2026-06-24
virtual fluid volume api 1.21.1

vinibad:
### Description
Currently, when mods (like the Create mod) render custom multiblock fluid tanks, they often bypass the standard block fluid rendering pipeline. This causes issues with modern shader packs that rely on accurate fluid boundaries to render volumetric underwater fog and calculate depth correctly. Often, this leads to the "white world" bug or completely invisible water surfaces.

This PR introduces the **Virtual Fluid Volume API** to solve this. It provides a way for mod developers to explicitly register custom, non-block-aligned fluid boundaries directly into Iris's rendering pipeline.

### Changes Made
* Added `VirtualFluidRegistry` and `VirtualFluidProvider` to allow mods to register callbacks for dynamic virtual fluid generation.
* Created the `VirtualFluidVolume` interface, allowing mods to define the geometry (quads), color, texture, and light properties of their fluid boundaries.
* Updated `VirtualFluidRenderer` to process and dispatch these registered virtual fluids during the translucent rendering phase.
* Ensured correct depth writes and shader uniform bindings (e.g., volumetric fog support) for virtual fluids.
* Added `DebugVirtualFluidVolume` built-in implementation. You can test it by launching the game with `-Diris.debugVirtualFluid=true`.

### Usage Example (Attached)
I am attaching a sample mod [CreateWatertanktest.zip](https://github.com/user-attachments/files/29272157/CreateWatertanktest.zip) that serves as a proof of concept. It uses a Mixin to cancel the default Create Mod fluid tank rendering and instead passes the tank's water coordinates directly to the new `VirtualFluidRegistry`.
When using a shader like Bliss Shaders, the water inside the tank is rendered flawlessly, including underwater volumetric fog and correct face culling, completely bypassing the visual glitches previously caused by forced buffer batches.

### Testing
* Verified that the debug virtual fluid renders correctly on all faces with proper culling.
* Verified that modded fluid rendering via the API correctly triggers volume
tric fog in shader packs without breaking the rendering state of the world.

### DIFF
--- a/common/src/api/java/net/irisshaders/iris/api/v0/IrisApi.java
+++ b/common/src/api/java/net/irisshaders/iris/api/v0/IrisApi.java
@@ -3,6 +3,8 @@ package net.irisshaders.iris.api.v0;
 import java.nio.ByteBuffer;
 import java.util.function.IntFunction;
 
+import net.irisshaders.iris.api.v0.virtualfluid.VirtualFluidProvider;
+
 /**
  * The entry point to the Iris API, major version 0. This is currently the latest
  * version of the API.
@@ -23,7 +25,7 @@ public interface IrisApi {
 	 * if they wish to check whether given API calls are available on
 	 * the currently installed Iris version.
 	 *
-	 * @return The current minor revision. Currently, revision 2.
+	 * @return The current minor revision. Currently, revision 3.
 	 */
 	int getMinorApiRevision();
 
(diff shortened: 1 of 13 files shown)


IMS212:
While AI generated PR’s aren’t entirely banned on this repo, PR’s where the dev has no clue what they made are.

this solution makes no logical sense as an API.

PR #3206, opened 2026-07-05
Fix inversed matrices for terrain

GeForceLegend:
Corrected name of inversed model view and projection matrix for sodium shader transformer, so terrain will receive correct values.

Notice: There is another issue on inversed projection matrix for hand. I'm not sure how to fix it so it's not included in this pr. Iris uses `CapturedRenderingState.INSTANCE.getGbufferProjection()` for inversed projection matrix, but projection matrix of this INSTANCE only sets one time per frame and it's value is global projection, so `gl_ProjectionMatrixInverse` for hand does not match inversed `gl_ProjectionMatrix`. Some related discussion can be found in https://discord.com/channels/774352792659820594/1522743919039877150


GeForceLegend:
Found some way to fix the hand projection inverse said above, but I'm not sure if it's a good way. Just provide it here as a reference.

[diff.patch](https://github.com/user-attachments/files/29668618/diff.patch)

PR #3211, opened 2026-07-06
Fix inversed projection matrix for hand

GeForceLegend:
Found a better way than [comment in #3206](https://github.com/IrisShaders/Iris/pull/3206#issuecomment-4884729485). If better way is found and applied, just close this.

### DIFF
--- a/common/src/main/java/net/irisshaders/iris/pathways/HandRenderer.java
+++ b/common/src/main/java/net/irisshaders/iris/pathways/HandRenderer.java
@@ -70,6 +70,7 @@ public class HandRenderer {
 		}
 		scaleMatrix.mul(poseStack.last().pose());
 		RenderSystem.setProjectionMatrix(cachedProjectionMatrixBuffer.getBuffer(scaleMatrix), ProjectionType.PERSPECTIVE);
+		CapturedRenderingState.INSTANCE.setGbufferProjection(scaleMatrix);
 
 		return new PoseStack();
 	}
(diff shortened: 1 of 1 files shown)


common/src/main/java/net/irisshaders/iris/pathways/HandRenderer.java
```diff
@@ -70,6 +70,7 @@ private PoseStack setupGlState(GameRenderer gameRenderer, CameraRenderState came
 		}
 		scaleMatrix.mul(poseStack.last().pose());
 		RenderSystem.setProjectionMatrix(cachedProjectionMatrixBuffer.getBuffer(scaleMatrix), ProjectionType.PERSPECTIVE);
+		CapturedRenderingState.INSTANCE.setGbufferProjection(scaleMatrix);
```
lowercasebtw, line 73: Just me thinking but wouldn't the solution to these just be injecting these methods in the respective methods

RenderSystem#setProjMatrix
RenderSystem#restoreProjMatrix ?

GeForceLegend: Iris is using projection matrix in `CapturedRenderingState.INSTANCE` for inversed projection [here](https://github.com/IrisShaders/Iris/blob/26.2/common/src/main/java/net/irisshaders/iris/pipeline/programs/ExtendedShader.java#L219). Unless Iris get projection matrix from RenderSystem (but currently there seems no easy way within vanilla code, it just store a gpu buffer and it may not being readable when trying to calculate projection inverse), we have to rely on something else for inversed special projection.

PR #3215, opened 2026-07-08
Fix Armourer's Workshop rendering compatibility

FortyTwoCn:
Fixes a rendering compatibility issue between Iris and Armourer's Workshop on NeoForge 1.21.1.
The previous `mergeRendering` flag was a global boolean. While Iris was rendering merged batches, every `RenderType.draw()` call could skip `setupRenderState()` / `clearRenderState()`, including render paths that were not part of the current Iris batch. This could leak render state into Armourer's Workshop GUI/item rendering, causing black quads, broken transparency, and in some modpacks a null shader crash.

Before：
image


Ru-Nan-i:
try [this](https://github.com/SAGESSE-CN/Armourers-Workshop/actions/runs/29164896253) version.

PR #3228, opened 2026-07-12
Wire up pipeline exclusion and add resource pack exclusion support

LLicorneuh:
While vanilla resource pack shaders are not currently a priority for Iris, we can still provide resource pack creators with the ability to disable specific rendering pipelines. This would allow certain core shaders to function correctly. I tested armor and entity shaders, and they already work quite well.

### DIFF
--- a/common/src/main/java/net/irisshaders/iris/Iris.java
+++ b/common/src/main/java/net/irisshaders/iris/Iris.java
@@ -236,6 +236,11 @@ public class Iris {
 			}
 		}
 
+		Minecraft minecraft = Minecraft.getInstance();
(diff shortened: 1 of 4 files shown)


IMS212:
This is not planned to be supported. This behavior forces otherwise correctly rendering objects to render to colortex0, which may not actually be the albedo buffer on many packs.  It is impossible to know what will actually be the albedo buffer, hence why this feature hasn't and will not be added.

This feature is guaranteed to break a pack's intended look, and in exchange, provides a slight chance of the RP shader looking only partially correct. I am not interested in allowing resource pack devs to intentionally break the behavior of any shader pack with a JSON.

PR #3295, opened 2026-08-15
Fix u_regionoffset being a required uniform

jbritain:
This often produces compile errors on shaderpacks which do not make use of the vertex position in a terrain shader.


GeForceLegend:
IMS already knows it here https://discord.com/channels/774352792659820594/774354522361299006/1534209595323711549

He may just haven't push local changes to github?

SpacEagle17:
@IMS212 no issues with this, right?

PR #3313, opened 2026-08-31
Fix support for 0.8.13

Wyvest:
Bumps mod version to 1.10.8 instead of .7, as the sodium breaks doesnt allow anything at or below .7


realduziy:
any eta on when this will be pushed?

PR #3320, opened 2026-09-07
Fix reverse z in gui

GeForceLegend:
This pr reverts changes in commit 82cd179, and replace it with a *better?* fix.

The reason of revert existing commmit is:
- This only considers vanilla item in gui. Gui from other mods that using depth is still incorrect, like xaero's minimap.
- This commit have some bug, which would let item in gui still bugged if launching game with shaders enabled, and then disable shaders after entering level.

This pr uses `ImmediateState.isRenderingLevel` to replace `ImmediateState.ALWAYS_REVERSE`, which should provide better compatibility for gui from other mods. The extra `Camera.update()` is required because this method will create the projection matrix used by sodium and then iris later, but it's called far away before isRenderingLevel set to true.

This pr will let item in gui works correctly, along side fix compatibility with xearo's minimap according to local test.


GeForceLegend:
Something that might be better applied. If you think previous is better, then just revert that.

